PR #33, opened 2019-05-16
Move Host resolver to Micronaut http-server module

sdelamo:
See: https://github.com/micronaut-projects/micronaut-core/pull/1675


jameskleeh:
Merged manually

PR #43, opened 2019-05-17
Add support for Graal 19.0.0

ilopmar:


jameskleeh:
@ilopmar Do we need anything else for the oauth2 module?

ilopmar:
I'd say no because oauth2 depends on jwt so everything should work. In any case, if you have a sample app that uses oauth2 I can convert it to native-image and check it.

jameskleeh:
@ilopmar Oauth2 has an optional dependency to JWT. It's only required to use OpenID providers. I'll drop you a sample app

PR #46, opened 2019-05-22
Add support a for default OAuth provider

jasondlee:
* Add configuration property
* Create default login and callback routes if a default provider is
  configured


jameskleeh:
Thanks for your contribution!

PR #48, opened 2019-05-26
Fix README title (s/RabbitMQ/Security/)

jchambers:
Looks like the README for Micronaut Security was copied from Micronaut RabbitMQ as a template, but the title wasn't updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Micronaut RabbitMQ
+# Micronaut Security
 
 [![Build Status](https://travis-ci.org/micronaut-projects/micronaut-security.svg?branch=master)](https://travis-ci.org/micronaut-projects/micronaut-security)
 


jameskleeh:
@jchambers Thanks! Can you sign the CLA?

jchambers:
So here's the thing: the bit of the CLA that deals with patent stuff prooooobably needs to get cleared by my employer's legal department for conflicts with our own IP assignment stuff. I'm sure it's not a problem, but it does mean taking their time and doing meetings and all that jazz. Given the size/significance of this contribution, I don't think that would be a productive use of anybody's time.

I don't mean to be difficult, but no, I don't think signing the CLA makes sense in the context of this particular contribution. Given something more substantial, I'd be happy to go through the process.

What's the best way forward? I pinky-swear I won't sue anybody if you merge this, but I can also close this pull request and open an issue in its place that points out the typo so somebody who _has_ signed the CLA can fix it.

jameskleeh:
@jchambers I just went ahead and fixed it. Thanks for the PR

jchambers:
Awesome. Thanks for understanding!

PR #58, opened 2019-07-20
Update Javadoc to make "user" not male-specific.

adityasrini:
"User" can be a machine, woman, man, multiple entities etc. This PR removes verbiage I came across while reading the docs that specifies the user as a man.

I hope the maintainers of this project understand implicit bias and appreciate PRs like this in the hopes of making the Micronaut community inclusive of all kinds of Engineers.


jameskleeh:
Thank you for the contribution!

PR #69, opened 2019-08-23
Make it possible to inject OpenIdProviderMetadata

lightoze:
In the old commit `@EachBean` was changed to `@Prototype` with "bean loop fix" comment but it contained a lot of other changes as well. At least nowadays tests are passing and having access to `OpenIdProviderMetadata` is very much needed because it contains configuration resolved from issuer url.


jameskleeh:
You can get the data from the client, however I merged this manually as it needed more changes

PR #71, opened 2019-08-29
Oauth grant classes cleanup, added refresh token request

lightoze:
This patch does some cleanup in OAuth grant classes and also ground work for refresh token handling.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/grants/RefreshTokenGrant.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/grants/RefreshTokenGrant.java
@@ -18,9 +18,13 @@ package io.micronaut.security.oauth2.grants;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.util.StringUtils;
+import io.micronaut.security.oauth2.configuration.OauthClientConfiguration;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+
+import java.util.Map;
 
 /**
  * Refresh Token Grant.
@@ -31,7 +35,11 @@ import edu.umd.cs.findbugs.annotations.Nullable;
  */
 @Introspected
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
-public class RefreshTokenGrant {
+public class RefreshTokenGrant extends AbstractSecureGrant implements AsMap {
+
+    private static final String KEY_GRANT_TYPE = "grant_type";
+    private static final String KEY_REFRESH_TOKEN = "refresh_token";
+    private static final String KEY_SCOPE = "scope";
 
     private String grantType = GrantType.REFRESH_TOKEN.toString();
     private String refreshToken;
@@ -41,12 +49,23 @@ public class RefreshTokenGrant {
      * Default constructor.
      */
     public RefreshTokenGrant() {
+    }
 
+    /**
+     * @param refreshToken The refresh token
+     * @param clientConfiguration The client configuration
+     * @since 2.0.0
+     */
+    public RefreshTokenGrant(String refreshToken, OauthClientConfiguration clientConfiguration) {
+        this.refreshToken = refreshToken;
+        scope = clientConfiguration.getScopes().stream()
+                .reduce((a, b) -> a + StringUtils.SPACE + b)
+                .orElse(null);
     }
 
     /**
      *
-     * @return refresh_token
+     * @return {@code "refresh_token"}
      */
     @NonNull
     public String getGrantType() {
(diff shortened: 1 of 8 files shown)


jameskleeh:
There are a few issues here:

* Why was AuthorizationCodeGrant and PasswordGrant changed?
  * Passing the client id and secret for authentication is a valid scheme and must be supported
* You can't change the public variable name in ClientCredentialsGrant. That is a breaking change.
* Is there a specification that says multiple scopes should be space separated for the client credentials grant? 
* OpenIdRefreshTokenRequestContext, OauthRefreshTokenRequestContext
  * Need the @author and @since tags modified to represent reality. Java doc needs updated

lightoze:
> * Why was AuthorizationCodeGrant and PasswordGrant changed?
>   * Passing the client id and secret for authentication is a valid scheme and must be supported

I removed code which was not used and didn't seem logical. Due to form serialization grant classes are not suitable as is for using them as request object so you have to convert them to `SecureGrantMap`. In terms of `TokenRequestContext`, that map **is** the grant, so **it** should implement `SecureGrant`.

> * You can't change the public variable name in ClientCredentialsGrant. That is a breaking change.

Valid point, but I doubt anyone is using it. Also I just now noticed that its `toMap` method returns normal hash map instead of `SecureGrantMap` which means client id/secret won't be sent, so if anyone would attempt using it it likely wouldn't work.

> * Is there a specification that says multiple scopes should be space separated for the client credentials grant?

Yes. https://tools.ietf.org/html/rfc6749#section-4.4.2

> * OpenIdRefreshTokenRequestContext, OauthRefreshTokenRequestContext
>   * Need the @author and @SInCE tags modified to represent reality. Java doc needs updated

Will fix. Should it be `@since 1.3`? I'll wait for your replies about all topics and make an update.

jameskleeh:
You may be right about the id/secret not working in its current form, however that does not mean the code should be removed without replacement. Yes this should be `@since 1.3.0`

lightoze:
Well the code that cannot and should not be used :)

lightoze:
I've made changes as discussed except for `ClientCredentialsGrant` - after reading the code it was broken at so many points (wrong value of constant in question, not returning `SecureMap` and thus unable to use client authentication) that I instead made even more incompatible change and changed constants to be private - in line with other grant types. I hope you understand.

`SecureGrant` is now implemented by new abstract class. When micronaut-projects/micronaut-core#1853 is fixed, all `toMap` methods can be deprecated and grant objects can be used directly.

lightoze:
@jameskleeh Is there anything yet to be resolved before this can be merged?

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/context/OpenIdCodeTokenRequestContext.java
```diff
@@ -56,19 +53,6 @@ public OpenIdCodeTokenRequestContext(AuthorizationResponse authorizationResponse
         this.oauthRouteUrlBuilder = oauthRouteUrlBuilder;
     }
 
-    /**
-     * Resolves the media type for the request body
-     *
-     * @param clientConfiguration The client configuration
-     * @return The media type
-     */
-    protected static MediaType getMediaType(OauthClientConfiguration clientConfiguration) {
```
jameskleeh: Moving this to `AbstractTokenRequestContext` doesn't make sense because this logic is specific to openid

security-oauth2/src/main/java/io/micronaut/security/oauth2/grants/RefreshTokenGrant.java
```diff
@@ -32,7 +35,11 @@
  */
 @Introspected
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
-public class RefreshTokenGrant {
+public class RefreshTokenGrant extends AbstractSecureGrant implements AsMap {
```
jameskleeh: Where does the spec say that the refresh token grant needs secured?

jameskleeh:
@lightoze See my comments

security-oauth2/src/main/java/io/micronaut/security/oauth2/grants/RefreshTokenGrant.java
lightoze: https://tools.ietf.org/html/rfc6749#section-6
>    Because refresh tokens are typically long-lasting credentials used to
   request additional access tokens, the refresh token is bound to the
   client to which it was issued.  If the client type is confidential or
   the client was issued client credentials (or assigned other
   authentication requirements), the client MUST authenticate with the
   authorization server as described in Section 3.2.1.

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/context/OpenIdCodeTokenRequestContext.java
lightoze: Fixed in new commit.

lightoze:
@jameskleeh I consolidated previous changes and rebased everything onto master. Should be good to merge before 2.0.0.

Issue #96, opened 2019-11-19
Make JWT Cookie login/logout redirect optional


mickoallen:
Can the redirect on successful/failed cookie login attempts be configurable? I'm in a situation where it would be preferable for my front end to handle navigation, post login.

mickoallen:
Something along these lines -
https://github.com/micronaut-projects/micronaut-security/pull/97

mickoallen:
added details to pr

PR #97, opened 2019-11-20
Issue #96 Jwt cookie no redirect

mickoallen:
Just feeling out if something like this would work out. Will do clean it up and add tests and documentation if we move forward.


alvarosanchez:
@mickoallen first, thank you for your contribution. Please sign the CLA if you want to move this forward.

Regarding the changes themselves, would you please clare to elaborate the issue you are trying to solve?

lehotskysamuel:
Hi, I'd also like to have an option to disable redirects. We use micronaut solely for our APIs and all redirection and routing logic is handled by react frontend. We have no endpoint that could be used for redirection. This applies both to login and logout.

mickoallen:
@alvarosanchez - similar as in @lehotskysamuel situation. It is common for the redirecting and routing to be handled by the UI and not redirects, that was the spirit behind this change. I'm happy to resurrect the pr and finish it off if it is something people want.

sdelamo:
With Micronaut Security class 2.0, we have an abstract class [CookieLoginHandler](https://github.com/micronaut-projects/micronaut-security/blob/master/security-jwt/src/main/java/io/micronaut/security/token/jwt/cookie/CookieLoginHandler.java) which exposes a couple of methods: 

- `public MutableHttpResponse<?> loginFailed(AuthenticationResponse authenticationFailed, HttpRequest<?> request)`
- `protected MutableHttpResponse<?> createSuccessResponse(HttpRequest<?> request)`
-  `protected MutableHttpResponse<?> createRefreshResponse(HttpRequest<?> request)`

I think that adding this class to your app will achieve what you are looking for:

```java
package io.micronaut.security.token.jwt.cookie;

import edu.umd.cs.findbugs.annotations.Nullable;
import io.micronaut.context.annotation.Replaces;
import io.micronaut.context.annotation.Requires;
import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.MutableHttpResponse;
import io.micronaut.security.config.RedirectConfiguration;
import io.micronaut.security.config.SecurityConfigurationProperties;
import io.micronaut.security.errors.PriorToLoginPersistence;
import io.micronaut.security.token.jwt.generator.AccessRefreshTokenGenerator;
import io.micronaut.security.token.jwt.generator.AccessTokenConfiguration;

import javax.inject.Singleton;

@Requires(property = SecurityConfigurationProperties.PREFIX + ".authentication", value = "cookie")
@Singleton
@Replaces(JwtCookieLoginHandler.class)
public class JwtCookieNonRedirectingLoginHandler extends JwtCookieLoginHandler {
    public JwtCookieNonRedirectingLoginHandler(RedirectConfiguration redirectConfiguration,
                                               JwtCookieConfiguration jwtCookieConfiguration,
                                               AccessTokenConfiguration accessTokenConfiguration,
                                               AccessRefreshTokenGenerator accessRefreshTokenGenerator,
                                               @Nullable PriorToLoginPersistence priorToLoginPersistence) {
        super(redirectConfiguration, jwtCookieConfiguration, accessTokenConfiguration, accessRefreshTokenGenerator, priorToLoginPersistence);
    }


    @Override
    protected MutableHttpResponse<?> createSuccessResponse(HttpRequest<?> request) {
        return HttpResponse.ok();
    }
}

```

@mickoallen I am closing this PR but let me know if you think the framework should behave in a different way.

sdelamo:
@alejandropg please, create a separate issue and lets discuss it there. Describe what you would like to achieve and change.

PR #118, opened 2019-11-29
Update securityRule.adoc

beedub:


alvarosanchez:
Fulfil is correct: https://www.grammarly.com/blog/fulfil-fulfill/

beedub:
Learned something new today :)

Issue #110, opened 2019-11-25
Crash when setting openid provider for Keycloak 8.0.0


TheoMuffin:
When setting openid issuer to that of Keycloak 8.0.0 then Micronaut crashes with ```Error instantiating bean of type [io.micronaut.security.oauth2.client.DefaultOpenIdClient]: No enum constant io.micronaut.security.oauth2.endpoint.AuthenticationMethod.TLS_CLIENT_AUTH```. The behavior doesn't change with different versions of Micronaut/security/oauth libraries, including snapshot versions. The problem goes away when using Keycloak 7.0.1 or older.

Steps to reproduce:

- Make a new bat file in Keycloak 8.0.0 > bin repository in order to start it up on a different port than micronaut: ```./standalone.bat -Djboss.socket.binding.port-offset=100```
- Configuring the first administration account at http://localhost:8180/auth/ may or may not be needed.
- Set the following in Micronauts application.yml: 
```
 security:
    enabled: true
    oauth2:
      enabled: true
      clients:
        fileserver:
          openid:
            issuer: 'http://localhost:8180/auth/realms/master'
    token:
      jwt:
        enabled: true
```
- Run micronaut

meamZ:
Can confirm this. Also if you just go through the steps at https://guides.micronaut.io/micronaut-oauth2-okta/guide/index.html but use Keycloak instead of Okta, you get the same issue.

goeh:
I can also confirm this. Keycloak 8.0.0 responds with `..., "token_endpoint_auth_methods_supported":["private_key_jwt","client_secret_basic","client_secret_post","tls_client_auth","client_secret_jwt"]` and this leads to exception in micronaut:
```
Error instantiating bean of type [io.micronaut.security.oauth2.client.DefaultOpenIdClient]: No enum constant io.micronaut.security.oauth2.endpoint.AuthenticationMethod.TLS_CLIENT_AUTH
                               |        at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1626)
                               |        at io.micronaut.context.DefaultBeanContext.createAndRegisterSingleton(DefaultBeanContext.java:2307)
                               |        at io.micronaut.context.DefaultBeanContext.getBeanForDefinition(DefaultBeanContext.java:1989)
                               |        at io.micronaut.context.DefaultBeanContext.getBeanInternal(DefaultBeanContext.java:1963)
[...]

PR #126, opened 2019-12-11
Fix for issue #110

istvanszoboszlai:
If an unknown authentication method is provided in the .well-known configuration Micronaut won't crash.


alvarosanchez:
Sorry, your PR does not work. It filters everything out. There is a failing test because of that.

Also, the issue was already in progress and has been fixed already.

Thanks anyway for your contribution!

istvanszoboszlai:
Sorry for causing unnecessary work for you guys. 
Next time i will try to test more locally. 
Actually it is interesting that it doesn't work. We use it locally, I'll check exactly why, just for my own piece of mind :)

And thanks for taking care of the core issue!

> Sorry, your PR does not work. It filters everything out. There is a failing test because of that.
> 
> Also, the issue was already in progress and has been fixed already.
> 
> Thanks anyway for your contribution!

Issue #108, opened 2019-11-22
io.micronaut.security.authentication.Authentication breaks the Serializable contract


stepanv:
# Environment
Micronaut Security 1.2.2

# Description
the `io.micronaut.security.authentication.Authentication` interface extends `Serializable`. In the same time, it declares a method
```
Map<String, Object> getAttributes();
```
if I understand the purpose of the attributes, even the `attributes` should be *serializable*, however, an `Object` may not be `Serializable`.

To implement this interface correctly, if I'm right (at least this is what works for me) one needs to have a field in the subtype:
```
HashMap<String, ? extends Serializable> attributes;
```
and then, when implementing the getter, the `attributes` need to be class-cast.

# Expected design
The `Authentication` interface should either
1. not extend the `Serializable` interface or
2. declare a method `Map<String, ? extends Serializable> getAttributes()`
so that one can implement the interface "correctly"

The current Micronaut implementations of the `Authentication` do break the Serializable contract (see https://rules.sonarsource.com/java/RSPEC-1948 )
(the `io.micronaut.security.authentication.UserDetails` violates the contract, the same is for `DefaultAuthentication`)

alvarosanchez:
@stepanv good point. Willing to contribute the change?

stepanv:
@alvarosanchez I'm definitely willing to contribute, but to make this right, it requires a design decision that will very likely break backwards compatibility. 
At least both options I'm suggesting are like that. 
For example, the PR I'm submitting is implementing the option [1].
Not sure what the "serializability" of the `Authentication` is for (honestly) but if it's really needed because of some framework/use-case you're aiming to support, the option [2] could be the way to go.
I can also imagine the problem can be solved by mentioning in the documentation that only Serializable values in the attributes map are guarantied to remain.

alvarosanchez:
Removing `Serializable` from `Authentication` is a breaking change that could only be done in an upcoming major release (2.x).

Changing `getAttributes()` return type is also breaking, and unfortunately can't be solved by introducing a `default` variant.

The only option I can imagine is to `@Deprecate` `Authentication` and replace it with a whole different interface. @jameskleeh thoughts?

jameskleeh:
It's not possible to remove Serializable from `Authentication`. The authentication objects need to support being serialized as part of session support. We can change the map and review any other classes in regards to serializability for 2.0

PR #128, opened 2019-12-12
Removing Serializable contract from Authentication

stepanv:
https://github.com/micronaut-projects/micronaut-security/issues/108 : removing 'Serializable' from Authentication to avoid breaking the Serialization contract.


jameskleeh:
The authentication needs to be serializable

PR #144, opened 2020-01-10
Make 'sub' claim optional, as per RFC 7519.

### LINKED ISSUE #135: JWT Requires optional field "sub"
While attempting to follow a [short example](https://medium.com/@thegiraffeclub/jwks-authentication-with-micronaut-bf28845d3b61) on using JWKS based JWT auth, I ran into this error:

```
21:11:05.021 [nioEventLoopGroup-1-5] DEBUG i.m.s.t.j.v.SubjectNotNullJwtClaimsValidator - JWT must contain a subject ('sub' claim)
21:11:05.022 [nioEventLoopGroup-1-5] DEBUG i.m.s.t.b.BasicAuthTokenValidator - Error while trying to Base 64 decode: eyJhbG
```
Based on RFC-7519 (JWT standard), this is not the case. This field is optional. It appears this is handled [here](https://github.com/micronaut-projects/micronaut-security/blob/master/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/SubjectNotNullJwtClaimsValidator.java). Should this restriction be relaxed?

alvarosanchez:
Fixes #135.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/SubjectNotNullJwtClaimsValidator.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/SubjectNotNullJwtClaimsValidator.java
@@ -30,7 +30,7 @@ import javax.inject.Singleton;
  * @since 1.1.0
  */
 @Singleton
-@Requires(property = JwtClaimsValidator.PREFIX + ".subject-not-null", notEquals = StringUtils.FALSE)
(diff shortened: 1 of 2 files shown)


sdelamo:
I don't think we should do this change. See comment in linked issue.

alvarosanchez:
I made this change as a fix for an incorrectly implemented specification. IMHO, it was a bug.

Anyway, there is no branch yet for version 2.0, so I will close this PR, which can be reopened once master becomes 2.x

PR #145, opened 2020-01-10
Allows disabling token propagation override.

### LINKED ISSUE #84: Token propagation should't override existing token in clients
Authorization header is getting unexpectedly overridden by the token propagation.

alvarosanchez:
Fixes #84


security/src/main/java/io/micronaut/security/token/propagation/TokenPropagationHttpClientFilter.java
```diff
@@ -92,11 +112,25 @@ public TokenPropagationHttpClientFilter(TokenWriter tokenWriter,
...
         return chain.proceed(targetRequest);
     }
+
+    private boolean hasExistingToken(MutableHttpRequest<?> targetRequest) {
+        String headerName = HttpHeaders.AUTHORIZATION;
+        if (httpHeaderTokenWriterConfiguration != null) {
+            headerName = httpHeaderTokenWriterConfiguration.getHeaderName();
+        }
+        return targetRequest.getHeaders().get(headerName) != null;
+    }
```
jameskleeh: This assumes the token will be written via headers. It should be the responsibility of the token writer to determine if the token already exists.

jameskleeh:
Also this is a new feature so this should target 1.3

security/src/main/java/io/micronaut/security/token/propagation/TokenPropagationHttpClientFilter.java
alvarosanchez: Good point, but actually now you mention, it's `TokenReader` the one that should determine whether the token exists

alvarosanchez:
Targeted master and used `TokenReader` to find the token.

The build will fail because master is broken.

jameskleeh:
Manually applied these changes to 1.4.x without a setting

PR #178, opened 2020-02-21
first validate signature with config alg supported

sdelamo:
After this change: https://github.com/micronaut-projects/micronaut-security/pull/95

We are attempting to validate a signed JWT against every signature configuration. It would be more efficient to attempt to validate the signature against signature configuration which declare themselves as compatible.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidatorUtils.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidatorUtils.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import java.text.ParseException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidatorUtils.java
```diff
@@ -88,22 +89,21 @@ private JwtTokenValidatorUtils() {
...
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("JWT Signature verification failed: {}", signedJWT.getParsedString());
-                    }
-                }
-            } catch (final JOSEException e) {
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("Verification fails with signature configuration: {}, passing to the next one", config);
-                }
+        }
+        for (final SignatureConfiguration config : signatureConfigurations.stream()
```
jameskleeh: Does this make sense? Would a signature configuration that doesn't support the algorithm every verify the jwt?

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidatorUtils.java
```diff
@@ -112,6 +112,34 @@ private JwtTokenValidatorUtils() {
...
 
+    /**
+     *
+     * Validates a Signed JWT signature against a signature configuration.
+     *
+     * @param signedJWT a Signed JWT Token
+     * @param config The signature configuration
+     * @return empty if signature validation fails
+     */
+    public static Optional<JWT> verifySignedJWTWithSignatureConfiguration(SignatureConfiguration config, SignedJWT signedJWT) {
```
jameskleeh: Make this private

sdelamo: I think it is valuable to have it public to allow people to call it.

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidatorUtils.java
sdelamo: Originally, we only verified JWT signatures with signature configurations which appear to be compatible (`SignatureConfiguration::supports(_) >> true`).

 What was reported with https://github.com/micronaut-projects/micronaut-security/pull/95 is that there are some tokens which don't specify any algorithm in the header. Thus, we did not attempt to validate them against any signature configuration. For those JWTs the only way to ensure if there is a valid configuration is to try to verify it directly against the signature configuration. 

What this PR aims to achieve is to try to verify first with the signature configurations which support the specified algorithm and if no verification is possible try the rest.

PR #196, opened 2020-03-16
Add support for GraalVM

ilopmar:


graemerocher:
Thinking about this I think it should be in core as that is where caffeine lives

ilopmar:
Which submodule and package should I use? `runtime` and `io.micronaut.cache` package?

ilopmar:
Sorry, there is already a `io.micronaut.cache.graal` package there

ilopmar:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/2931

PR #212, opened 2020-04-07
DefaultSecurityService.hasRole honor the custom roles-name key Fixes #211

### LINKED ISSUE #211: DefaultSecurityService.hasRole(String roleName) should honor custom roles-name setting of token
The DefaultSecurityService.hasRole(String roleName) currently by default searches the the "roles" attribute of token. This method should honor the custom settings set in property micronaut.security.service.token.roles-name

### Steps to Reproduce

1. Set custom roles-name property
2. Attempt to call securityService.hasRole("roleName")

### Expected Behaviour

The role should be found

### Actual Behaviour

Role is not found

mathwizard:


sdelamo:
Thanks for the @PR @mathwizard I will do some small changes and merge this shortly.

sdelamo:
Close by: https://github.com/micronaut-projects/micronaut-security/pull/212

PR #237, opened 2020-05-15
refactor: simpler refresh token generator

sdelamo:
This is an attempt to simplify `refresh token` generation. 

This is our current implementation

https://github.com/micronaut-projects/micronaut-security/blob/5eab9f801fb43a11c7575078b83be7c648c84ac7/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/SignedRefreshTokenGenerator.java

and this is my proposed solution: 

https://github.com/micronaut-projects/micronaut-security/blob/simpler-refresh-token-generator/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/JwsRefreshTokenGenerator.java

It leverages [Nimbus JOSE](https://connect2id.com/products/nimbus-jose-jwt/examples/jws-with-hmac), which is already in the classpath, to sign a payload. 

Moreover, it has more configuration options. For example, it enables you to configure a different JWS Algorithm than default HS256.

It should be more familiar to Micronaut users since it is similar configuration than our [SecretSignatureConfiguration](https://github.com/micronaut-projects/micronaut-security/blob/master/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/secret/SecretSignatureConfiguration.java) used to sign JWT. 

Moreover, this PR removes the need to enable `refresh-token`. 

Instead of having:

```
micronaut:
  security:
    token:
      jwt:
        generator:
          refresh-token:
            enabled: true 
            secret: pleaseChangeThisSecretForANewOne 
```

it will suffice to have: 

```
micronaut:
  security:
    token:
      jwt:
        generator:
          refresh-token:
            secret: pleaseChangeThisSecretForANewOne 
```

That it is to say. When you have a bean of type `RefreshTokenGenerator`, `RefreshTokenValidator`, `RefreshTokenPersistence`, `AccessRefreshTokenGenerator` generates a refresh-token.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/AccessRefreshTokenGenerator.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/AccessRefreshTokenGenerator.java
@@ -15,7 +15,9 @@
  */
 package io.micronaut.security.token.jwt.generator;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.context.BeanContext;
 import io.micronaut.context.event.ApplicationEventPublisher;
 import io.micronaut.security.authentication.UserDetails;
 import io.micronaut.security.token.generator.RefreshTokenGenerator;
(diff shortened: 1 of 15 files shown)


sdelamo:
Please, note I added a test to master branch to verify a refresh token is not able to be used as an access token : 

https://github.com/micronaut-projects/micronaut-security/commit/72e4d0de01080886ce2e9963388d52bf1b4666f1

sdelamo:
@jameskleeh this should be ready for review.

sdelamo:
I am going to ahead and merge this. It is encapsulated in a single commit.

Issue #252, opened 2020-05-23
Authorization header prefix treated case-sensitive


jcassee:
The implementation of `HttpHeaderTokenReader.extractTokenFromAuthorization` [treats the header value prefix as case sensitive](https://github.com/micronaut-projects/micronaut-security/blob/987e3d3c1962c1152dc74eaeed0bdba2e48cbc81/security/src/main/java/io/micronaut/security/token/reader/HttpHeaderTokenReader.java#L74). Although [the OAuth spec seems to suggest that is it](https://tools.ietf.org/html/rfc6750#section-2.1), [the HTTP spec says that this prefix is a case insensitive token](https://tools.ietf.org/html/rfc7235#section-2.1).

Can the check for the prefix be made to be case-insensitive? (I can submit a PR is that is useful.)

sdelamo:
We currently support: 

`Bearer xx.yy.zz => xx.yy.zz `

and you want to support as well:

`bearer xx.yy.zz => xx.yy.zz`

right?

sdelamo:
And of course PRs are welcomed.

jcassee:
@sdelamo Yes, all lower-case is my immediate use case, but "BeArEr" is also valid.

I'll create a PR. I see there are no immediate tests for this class. Do you want me to create one? If so, what would be a good location?

sdelamo:
> I'll create a PR. I see there are no unit tests for this class. Do you want me to create a test? If so, what would be a good location?

You will see many places int he framework where we don't have unit tests but the class is tested in a functional tests which engages the bean. 

However, it is great that you added the functional test.

PR #257, opened 2020-05-26
[WIP] Make Authorization header prefix check case insensitive

jcassee:
See #252

### DIFF
--- a/security/src/main/java/io/micronaut/security/token/reader/HttpHeaderTokenReader.java
+++ b/security/src/main/java/io/micronaut/security/token/reader/HttpHeaderTokenReader.java
@@ -71,7 +71,7 @@ public abstract class HttpHeaderTokenReader implements TokenReader {
             sb.append(" ");
         }
         String str = sb.toString();
-        if (authorization.startsWith(str)) {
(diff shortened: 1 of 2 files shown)


graemerocher:
Maybe the change should include `toLowerCase(Locale.ENGLISH)` as it could break depending on the system locale?

jcassee:
> Maybe the change should include `toLowerCase(Locale.ENGLISH)` as it could break depending on the system locale?

Or `Locale.ROOT`? A quick "find usages" in my IDE finds many similar case-insensitive checks around HTTP are using it.

jcassee:
Question: is using a stub the preferred way of testing an abstract class? The Spock docs mention using a partial mock using `Spy`, but from a quick search through the project code, I don't see `Spy` used anywhere.

sdelamo:
Thanks for the PR @jcassee 

> Question: is using a stub the preferred way of testing an abstract class? The Spock docs mention using a partial mock using `Spy`, but from a quick search through the project code, I don't see `Spy` used anywhere.

Yes, in general we don't use `Spy`.

PR #283, opened 2020-06-12
BREAKING CHANGE: delete UserDetails

sdelamo:
I think one of the more confusing parts of the security API is the use of `UserDetails`, an implementation, instead of `Authentication` the interface. 

The correct way to instantiate a `UserDetails` should be 

```java
new UserDetails('john', ['ROLE_USER'], [tokenConfiguration.getRolesName(): ['ROLE_USER'], [tokenConfiguration.getNameKey(): `john`)
```

and then we should be able to expose a `getAttributes()` method in `UserDetails`. 

Initially `UserDetails` did not have attributes. Moreover, the above instantiation is cumbersome. What we did is don't expose `getAttributes()` and expose 

```public Map<String, Object> getAttributes(String rolesKey, String usernameKey)``` instead. 

So, to go from `UserDetails` to `Authentication` you will do something like: 

```java
UserDetails userDetails = new UserDetails('john', ['ROLE_USER'])
new Authentication() {
   @Override
   String getName() {
       userDetails.getName()
   }
   
   Map<String, Object> getAttributes() {
      [tokenConfiguration.getRolesName(): userDetails.getRoles(), [tokenConfiguration.getNameKey(): userDetails.getUsername())
   }
}
```

which is a pretty hacky. 

This PR:

- Removes UserDetails and just use `Authentication`. 
- It provides several static methods to `AuthenticationResponse` and `Authentication` to ease the instances of those interfaces. 
- It replaces the use through the API of `UserDetails` with `Authentication`.
- It renames `io.micronaut.security.oauth2.endpoint.token.response.OauthUserDetailsMapper` to `io.micronaut.security.oauth2.endpoint.token.response.OauthAuthenticationMapper`
- It renames `io.micronaut.security.oauth2.endpoint.token.response.OpenIdUserDetailsMapper` to ` import io.micronaut.security.oauth2.endpoint.token.response.OpenIdAuthenticationMapper`.
- It renames `io.micronaut.security.oauth2.endpoint.token.response.DefaultOpenIdUserDetailsMapper` to`io.micronaut.security.oauth2.endpoint.token.response.DefaultOpenIdAuthenticationMapper`.


sdelamo:
UserDetails is causing issues

https://github.com/micronaut-projects/micronaut-security/issues/559

Issue #284, opened 2020-06-12
BasicAuthAuthenticationConfiguration defines its own roles key but TokenConfiguration should be used instead


sdelamo:
Steps to reproduce:

```java
package com.example;

import io.micronaut.http.MediaType;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;
import io.micronaut.http.annotation.Produces;

import javax.annotation.security.RolesAllowed;
import java.security.Principal;

@Controller
public class HomeController {

    @RolesAllowed("ROLE_ADMIN")
    @Produces(MediaType.TEXT_PLAIN)
    @Get
    String index(Principal principal) {
        return principal.getName();
    }
}
```

```
package example.micronaut;

import edu.umd.cs.findbugs.annotations.Nullable;
import io.micronaut.http.HttpRequest;
import io.micronaut.security.authentication.AuthenticationProvider;
import io.micronaut.security.authentication.AuthenticationRequest;
import io.micronaut.security.authentication.AuthenticationResponse;
import io.micronaut.security.authentication.UserDetails;
import io.reactivex.BackpressureStrategy;
import io.reactivex.Flowable;
import org.reactivestreams.Publisher;

import javax.inject.Singleton;
import java.util.Collections;

@Singleton
class CustomAuthProvider implements AuthenticationProvider {

    @Override
    public Publisher<AuthenticationResponse> authenticate(@Nullable HttpRequest<?> httpRequest, AuthenticationRequest<?, ?> authenticationRequest) {
        return Flowable.create(emitter -> {
            emitter.onNext(new UserDetails("admin", Collections.singletonList("ROLE_ADMIN")));
            emitter.onComplete();
        }, BackpressureStrategy.ERROR);
    }
}
```

`curl "http://localhost:8080"   -u 'admin:admin'` returns `admin`. 

If I set: 

```yaml
---
micronaut:
  security:
    basic-auth:
      roles-name: 'foo'
```

and restart the app 

`curl "http://localhost:8080"   -u 'admin:admin'` returns `403`. 

Micronaut Security 2.0.0.M3

PR #285, opened 2020-06-12
fix: BasicAuthAuthenticationFetcher should use TokenConfiguration

sdelamo:
Closes bug: https://github.com/micronaut-projects/micronaut-security/issues/284

We should always use `TokenConfiguration::getRolesNames`. 

Close: https://github.com/micronaut-projects/micronaut-security/issues/286 as well.


security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationFetcher.java
```diff
@@ -65,9 +67,9 @@ public BasicAuthAuthenticationFetcher(Authenticator authenticator,
             Flowable<AuthenticationResponse> authenticationResponse = Flowable.fromPublisher(authenticator.authenticate(request, credentials.get()));
 
             return authenticationResponse.switchMap(response -> {
-                if (response.isAuthenticated()) {
+                if (response.isAuthenticated() && response.getUserDetails().isPresent()) {
```
jameskleeh, line 70: The contract of the API is that if `isAuthenticated` returns true, the user details should be present. So this check is unnecessary

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationFetcher.java
```diff
@@ -36,21 +38,21 @@
...
-    private final BasicAuthAuthenticationConfiguration configuration;
+    private final TokenConfiguration configuration;
 
     /**
      * @param authenticator The authenticator to authenticate the credentials
      * @param configuration The basic authentication configuration
      */
     public BasicAuthAuthenticationFetcher(Authenticator authenticator,
-                                          BasicAuthAuthenticationConfiguration configuration) {
+                                          TokenConfiguration configuration) {
```
jameskleeh, line 55: I'm not sure I agree we should be using token configuration here. A basic auth string is not a token, so it doesn't make sense for users to configure it via ...token..

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationConfiguration.java
```diff
@@ -1,70 +0,0 @@
...
-import io.micronaut.security.config.SecurityConfigurationProperties;
-
-/**
- * Configuration for basic authentication.
- *
- * @author James Kleeh
- * @since 2.0.0
- */
-@ConfigurationProperties(BasicAuthAuthenticationConfiguration.PREFIX)
-public class BasicAuthAuthenticationConfiguration implements Toggleable {
```
jameskleeh, line 29: I don't think this should be deleted for documentation purposes

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationFetcher.java
sdelamo: Maybe the issue is `TokenConfiguration` should be renamed to `AuthenticationConfiguration`.

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationFetcher.java
sdelamo: I've removed the check.

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationConfiguration.java
sdelamo: I've documented in the docs how to disable. I am not confortable leaving a class just for the `Toggleable` field. Moreover, it was never displayed in the main docs. Only in configuration reference.

sdelamo:
@jameskleeh what do you think? Do you want me to rename `TokenConfiguration` to `AuthenticationConfiguration`?

security/src/main/java/io/micronaut/security/authentication/BasicAuthAuthenticationFetcher.java
sdelamo: Please note that the reason for this pr is that not using `TokenConfiguration::getRolesName` cuases a bug. See steps the reproduce here: https://github.com/micronaut-projects/micronaut-security/issues/284

PR #293, opened 2020-06-23
DefaultSecurityService should use RolesFinder

dstepanov:
Custom roles finder should provider roles for the security service, `DefaultRolesFinder` didn't check if tone configuration is enabled or not.

### DIFF
--- a/security/src/main/java/io/micronaut/security/token/DefaultRolesFinder.java
+++ b/security/src/main/java/io/micronaut/security/token/DefaultRolesFinder.java
@@ -18,6 +18,7 @@ package io.micronaut.security.token;
 import io.micronaut.security.token.config.TokenConfiguration;
 
 import javax.annotation.Nonnull;
+import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


traycho:
What if roles are not provided from the token ?

Having roles as part of Authentication just as an attribute (as it is right now)  sounds better for abstraction how roles are populated.

dstepanov:
@traycho There is already `RolesFinder` that is doing the extraction, the problem is if you provide custom impl. `SecurityService` is not going to pick up your new roles which is a bug.

dstepanov:
@traycho Check `DefaultRolesFinder`

dstepanov:
@graemerocher @jameskleeh Can somebody take a look at this one? We override `RolesFinder` but `SecurityService` is not using it which I think is not correct.

dstepanov:
@sdelamo Would you have time to review this pr?

sdelamo:
Thanks for the PR @dstepanov I will review this now.

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-security/pull/328

sdelamo:
@dstepanov The PR closing this will fix this for Micronaut Security 2.x Do you need this fix for 1.4.x. branch ?

PR #297, opened 2020-06-24
hasRole check should case insensitive validation in both ways for String and Collections.

traycho:
As software engineer I would expect roles checks to be case-insensitive.

This change will perform equalsIgnoreCase for both strings and collections of roles.

### DIFF
--- a/security/src/main/java/io/micronaut/security/utils/DefaultSecurityService.java
+++ b/security/src/main/java/io/micronaut/security/utils/DefaultSecurityService.java
@@ -37,7 +37,6 @@ public class DefaultSecurityService implements SecurityService {
     private final TokenConfiguration tokenConfiguration;
 
     /**
-     *
      * @param tokenConfiguration Token Configuration
      */
     public DefaultSecurityService(TokenConfiguration tokenConfiguration) {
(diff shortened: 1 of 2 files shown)


sdelamo:
Defining roles with `@Secured` annotation or with[ intercept url maps](https://micronaut-projects.github.io/micronaut-security/latest/guide/index.html#interceptUrlMap) is  is case sensitive. 

I think we already had tests for it but I have added an extra test to be clear about it: 

https://github.com/micronaut-projects/micronaut-security/commit/08fcd9d3c306055a737995465c94bdd7c79707a2

Thus,` SecurityService::hasRole `should be case sensitive both for collections and strings as well. Changing it to be case insensitive is a breaking change which we will not be able to add until the next mayor version.

sdelamo:
I've created an issue about this https://github.com/micronaut-projects/micronaut-security/issues/327

sdelamo:
You will be able to achieve case insensitive by overriding `RolesFinder`

```
@Singleton
@Replaces(DefaultRolesFinder)
public class RolesFinderReplacement extends DefaultRolesFinder {
    public RolesFinderReplacement(TokenConfiguration tokenConfiguration) {
        super(tokenConfiguration)
    }

    @Override
    public boolean hasAnyRequiredRoles(List<String> requiredRoles, List<String> grantedRoles) {
        for (String role : requiredRoles) {
            if (grantedRoles.stream().anyMatch(grantedRole -> grantedRole.equalsIgnoreCase(role))) {
                return true;
            }
        }
        return false;
    }
}
```

traycho:
What happens with having an option to switch without writing extra classes ?

sdelamo:
I have reopened this issue: https://github.com/micronaut-projects/micronaut-security/issues/327 

Maybe in the future we can make it configurable. 

Personally, I think that the current default is a good. Also, we have an easy workaround with one method override. Thus, I think the issue has a low priority.

PR #316, opened 2020-07-17
Check if the Authorization value is long enough to substring the prefix

swaechter:
The authentication fetcher `BasicAuthAuthenticationFetcher` strips away the leading `Basic ` prefix in order to get the Base64 encoded payload. The problem is that this substringing lacks a check that ensures that the header field has at least the length of the prefix (If not, `substring` will throw a `StringIndexOutOfBoundsException` exception that breaks the authentication flow).

Example:

`Authorization: ` --> StringIndexOutOfBoundsException
`Authorization: 123` --> StringIndexOutOfBoundsException
`Authorization: Basic` --> StringIndexOutOfBoundsException
`Authorization: Basic ` --> `Authorization: Basic` (Header values are trimmed) -->  StringIndexOutOfBoundsException
`Authorization: Basic <PROPERBASE64PAYLOAD>` --> Ok

This PR adds a check that ensure that the length of the header field is >= the prefix length. If not, an `Optional.empty` is passed down the chain and the following map/flatMaps actions are not executed/skipped


sdelamo:
Closed by: https://github.com/micronaut-projects/micronaut-security/pull/321

Issue #327, opened 2020-07-23
Roles should be case insensitive


sdelamo:
I think the best default is what we currently have. That it is case sensitive.

However, some users expect them to be case insensitive: 

[hasRole check should case insensitive validation in both ways for String and Collections](https://github.com/micronaut-projects/micronaut-security/pull/297). 

Maybe, we can make it configurable.

sdelamo:
You will be able to achieve case insensitive by overriding `RolesFinder`

```
@Singleton
@Replaces(DefaultRolesFinder)
public class RolesFinderReplacement extends DefaultRolesFinder {
    public RolesFinderReplacement(TokenConfiguration tokenConfiguration) {
        super(tokenConfiguration)
    }

    @Override
    public boolean hasAnyRequiredRoles(List<String> requiredRoles, List<String> grantedRoles) {
        for (String role : requiredRoles) {
            if (grantedRoles.stream().anyMatch(grantedRole -> grantedRole.equalsIgnoreCase(role))) {
                return true;
            }
        }
        return false;
    }
}
```

PR #289, opened 2020-06-17
Small refactoring of TokenAuthenticationFetcher

dstepanov:

PR #328, opened 2020-07-23
DefaultSecurityService should use RolesFinder

sdelamo:
I agree with you @dstepanov `DefaultSecurityService`should use `RolesFinder`. Could you check this PR and let me know if it aligns with what you are looking for. 

Aside: I've created a `RolesFinder::hasAnyRequiredRoles` so that we have a single place which can be override to support for example [Case Insensitive roles](https://github.com/micronaut-projects/micronaut-security/issues/327)


dstepanov:
Looks good, maybe it would be better to have `RolesFinder::hasAnyRequiredRoles(Claims claims, requiredRoles)`
so this code can be reduced to one call:
```
.map(rolesFinder::findInClaims)
.map(grantedRoles -> rolesFinder.hasAnyRequiredRoles(Collections.singletonList(role), grantedRoles))
```

BTW: Can you please take a look at https://github.com/micronaut-projects/micronaut-security/pull/289? I can redo PR for master.

sdelamo:
I've created a `RolesFinder::hasAnyRequiredRoles(List<String> requiredRoles, Authentication authentication)` to simplify things further.

Issue #346, opened 2020-08-15
JWT validation failed using Cognito with Google Oauth2


dgbmariano:
We are updating to Micronaut 2.0, but a problem occur with our current login workflow. We use cognito as our authentication provider and the login is made using Google.
When we try to sign in in the first time, we receive the following error:
```
{"message":"JWT validation failed","_links":{"self":{"href":"/oauth/callback/cognito?code=xxxxxx....","templated":false}}}
```
If we go back to the sign in page and try again the authentication proccess is successful.

### Steps to Reproduce

1. Clone this repo: https://github.com/dgbmariano/cognito-authentication-problem
2. Follow this guide to configure Cognito in AWS: https://guides.micronaut.io/micronaut-oauth2-cognito/guide/index.html
3. When modifying the app Settings, set this configurations:
![image](https://user-images.githubusercontent.com/55102164/90302073-1a7d0a00-de7a-11ea-9c74-ad71474dd7b7.png)
4. You will need to create a new Google OAuth 2.0 Client:
![image](https://user-images.githubusercontent.com/55102164/90302089-4d270280-de7a-11ea-8381-b4b1d63b7934.png)
6. Now replace the `client-id`, `client-secret` and the necessary fields in issuer in the `application.yml`.
7. **Run the application** and open http://localhost:8080/
8. Click in Enter
9. After sign in with Google, the authentication will fail with the message:
```
{"message":"JWT validation failed","_links":{"self":{"href":"/oauth/callback/cognito?code=xxxxxx....","templated":false}}}
```
10. Go back to http://localhost:8080/ and click in Enter again. The authentication will be successful now.

### Expected Behaviour

The authentication should be successful in the first try.

### Actual Behaviour

It fail once, then it authenticate.

### Environment Information

- **Operating System**: Ubuntu 20.02
- **Micronaut Version:** 2.0.1

sdelamo:
We include a nonce. More info about nonce validation https://openid.net/specs/openid-connect-core-1_0.html#IDToken

>  String value used to associate a Client session with an ID Token, and to mitigate replay attacks. 
> The value is passed through unmodified from the Authentication Request to the ID Token.
[...]

PR #357, opened 2020-08-18
refactor: extract two proteced methods to simplify override

sdelamo:
see: https://github.com/micronaut-projects/micronaut-security/issues/346


alvarosanchez (changes requested):
There are conflicts, not sure if the GitHub diff would be different once you resolve them

alvarosanchez (approved):

PR #361, opened 2020-08-20
build: bump up kotlin to 1.4.0

sdelamo:


ilopmar:
@sdelamo can we switch it to draft PR so we accidentally don't merge it?

sdelamo:
I am flagging them with `status: future consdieration`I am not against changing to a draft PR. I don't see the option in the UI.

PR #371, opened 2020-08-30
Support for passing token through URL parameters.

joaquim-manzini:
I've needed a way to be authenticated through SSE which doesn't allow headers.


jameskleeh:
The `BearerTokenReader` is responsible for reading the token from the `Authorization` header. If you want the token to be read from another location, a separate token reader should be created to do so. That wouldn't require any changes in this module to do. You can add it in your app

joaquim-manzini:
Well, thanks anyway.

Server Side Events doesn't allow headers. With this in mind I've created this option for my own use of SSE with Micronaut and it was successful.

I thought that it would be nice for the Micronaut platform. 

When we pass the token through URL query parameter it's still a bearer token. Isn't it? Grails Spring Security Rest does something [similar](http://alvarosanchez.github.io/grails-spring-security-rest/latest/docs/#_sending_tokens_in_the_request).

joaquim-manzini:
Thanks!! God bless!

PR #403, opened 2020-09-25
feat: add introspection endpoint

sdelamo:
[OAuth 2.0 Token Introspection](https://tools.ietf.org/html/rfc7662)


sdelamo:
@jameskleeh  I've added a GET /token_info as well. Both POST and GET /token_info require authentication

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-security/pull/405

PR #404, opened 2020-09-25
refactor: split DefaultOpenIdAuthorization methods

sdelamo:
The goal is that you can override easily the request being made to the token endpoint. If you are working with an authorization server non standard. 

e.g. You could just override the `sendRequest` method. 

```java
    protected Publisher<OpenIdTokenResponse> sendRequest(OpenIdAuthorizationResponse authorizationResponse,
                                                         OauthClientConfiguration clientConfiguration,
                                                         SecureEndpoint tokenEndpoint) {
        return httpClient.retrieve(createRequest(authorizationResponse, clientConfiguration), OpenIdTokenResponse.class);
    }

    private HttpRequest<?> createRequest(OpenIdAuthorizationResponse authorizationResponse,
                                      OauthClientConfiguration clientConfiguration) {
        return HttpRequest.POST("/api/v1/openid/token", new HashMap<String, String>() {{
            put("code", authorizationResponse.getCode());
            put("client_id", clientConfiguration.getClientId());
            put("grant_type", GrantType.AUTHORIZATION_CODE.toString());
        }}).contentType(MediaType.APPLICATION_FORM_URLENCODED)
                .accept(MediaType.APPLICATION_JSON)
                .basicAuth(clientConfiguration.getClientId(), clientConfiguration.getClientSecret());
    }
```

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/authorization/response/DefaultOpenIdAuthorizationResponseHandler.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/authorization/response/DefaultOpenIdAuthorizationResponseHandler.java
@@ -16,20 +16,25 @@
 package io.micronaut.security.oauth2.endpoint.authorization.response;
 
 import com.nimbusds.jwt.JWT;
+import edu.umd.cs.findbugs.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


sdelamo:
@jameskleeh I have made the methods private to keep the API surface the same. I think having more methods makes the class easier to work with.

PR #405, opened 2020-09-26
feat: add introspection endpoint

sdelamo:


graemerocher:
@sdelamo was this reviewed by someone?

sdelamo:
Yes, James review it.

graemerocher:
ok, if you can make that visible in terms of requesting a review in future since it seemed like it was not the case from the PR history

sdelamo:
> ok, if you can make that visible in terms of requesting a review in future since it seemed like it was not the case from the PR history

Will do.

PR #412, opened 2020-10-09
Remove NonNull annotation of refreshToken parameter

morki:
I think it is not valid, because `refreshToken` parameter can be `null` if `RefreshTokenValidator` and `RefreshTokenPersistence` beans are not provided.

Now it is hard to extend and override those methods, because in Kotlin it results in non-nullable type and invalid checking agains `null` value.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/DefaultAccessRefreshTokenGenerator.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/DefaultAccessRefreshTokenGenerator.java
@@ -164,7 +164,7 @@ public class DefaultAccessRefreshTokenGenerator implements AccessRefreshTokenGen
      */
     @NonNull
     @Override
-    public Optional<AccessRefreshToken> generate(String refreshToken, @NonNull UserDetails userDetails) {
+    public Optional<AccessRefreshToken> generate(@Nullable String refreshToken, @NonNull UserDetails userDetails) {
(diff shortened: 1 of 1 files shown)


sdelamo:
@morki can you change your PR to use `@Nullable` instead of removing `@NonNull`?

security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/DefaultAccessRefreshTokenGenerator.java
```diff
@@ -164,7 +164,7 @@ public DefaultAccessRefreshTokenGenerator(AccessTokenConfiguration accessTokenCo
      */
     @NonNull
     @Override
-    public Optional<AccessRefreshToken> generate(@NonNull String refreshToken, @NonNull UserDetails userDetails) {
+    public Optional<AccessRefreshToken> generate(String refreshToken, @NonNull UserDetails userDetails) {
```
sdelamo, line 167: can you add @Nullable here as well ?

sdelamo (changes requested):

security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/DefaultAccessRefreshTokenGenerator.java
morki: Of course, sorry I has overlooked it.

sdelamo (approved):

PR #415, opened 2020-10-15
Opt-out kotlin stdlib api dependency closes #414

### LINKED ISSUE #414: Micronaut Security is pulling in the kotlin standard lib
Same issue as in core, the Kotlin plugin now applys the standard lib as an API dependency. Has to be disabled with this property https://github.com/micronaut-projects/micronaut-core/blob/2.1.x/gradle.properties#L147

swanandvk:
Disable Kotlin stdlib as api dependency closes #414 using **gradle.properties**
`kotlin.stdlib.default.dependency=false`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,6 +20,7 @@ gebVersion=3.4
 seleniumVersion=3.141.59
 geckodriverVersion=0.26.0
 org.gradle.caching=true
+kotlin.stdlib.default.dependency=false
 # Needed for documentation generation. Needs to be Grails 3.2.x not 3.3.x
 grailsVersion=3.3.10
 testssecurity=security/src/test/groovy/io/micronaut/docs


sdelamo:
[Dependency on the standard library](https://kotlinlang.org/docs/reference/using-gradle.html#dependency-on-the-standard-library)

> A dependency on a standard library (stdlib) in each source set is added automatically. The version of the standard library is the same as the version of the Kotlin Gradle plugin.

> For platform-specific source sets, the corresponding platform-specific variant of the library is used, while a common standard library is added to the rest. The Kotlin Gradle plugin will select the appropriate JVM standard library depending on the kotlinOptions.jvmTarget compiler option of your Gradle build script.

> If you declare a standard library dependency explicitly (for example, if you need a different version), the Kotlin Gradle plugin won’t override it or add a second standard library.

> If you do not need a standard library at all, you can add the opt-out flag to the gradle.properties:

> `kotlin.stdlib.default.dependency=false`

sdelamo (approved):
Thanks a lot for the PR: @swanandvk 
I have added a blockquote to the Kotlin docs for reference.

PR #451, opened 2020-11-13
Update jwtGeneratorSignature.adoc

pratik-brex:


sdelamo:
Thanks @pratik-brex 

See https://micronaut-projects.github.io/micronaut-security/latest/guide/index.html#jwtGeneratorSignature

> To enable a JWT signature in token generation, you need to have in your app a bean of type SignatureGeneratorConfiguration qualified with name `generator`.

However, to verify a signature you don't need a name qualifier. Thus, I don't think your doc changes are correct.

pratik-brex:
> Thanks @pratik-brex
> 
> See https://micronaut-projects.github.io/micronaut-security/latest/guide/index.html#jwtGeneratorSignature
> 
> > To enable a JWT signature in token generation, you need to have in your app a bean of type SignatureGeneratorConfiguration qualified with name `generator`.
> 
> However, to verify a signature you don't need a name qualifier. Thus, I don't think your doc changes are correct.

Discussed this with @jameskleeh in the micronaut gitter channel. I couldn't get it to work without the `@Named` annotations

jameskleeh:
@sdelamo In the case of `@EachBean`, the target bean must be qualified

```
@EachBean(SecretSignatureConfiguration.class)
public class SecretSignature
```

PR #457, opened 2020-11-18
build: bump up to gradle 6.7.1

sdelamo:


alvarosanchez:
You shouldn't need PRs like this. The wrapper gets updated weekly on the template repo, and then propagated to all sync'ed repos

PR #463, opened 2020-11-19
Add refresh token request context

lightoze:
Here's a second try for #71 but with a more focused approach.

It's a bummer not being able to refresh access token from Micronaut OAuth2 client. This patch makes it easier to handle workflows where backend needs to access resources on behalf of the client in background. You only need to create a request context and then run it through `TokenEndpointClient` and `OpenIdTokenResponseValidator`, and you've got a new set of tokens!

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/context/OauthRefreshTokenRequestContext.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/context/OauthRefreshTokenRequestContext.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.security.oauth2.endpoint.token.request.context;
+
+import io.micronaut.core.type.Argument;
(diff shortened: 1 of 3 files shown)


lightoze:
Build seems to be broken for unrelated reason.

sdelamo:
@lightoze I will try to look into this PR this week or the next. Sorry for the delay.

ToniA:
Cool, I was just working on with Keycloak token refresh. I would also suggest adding property refreshExpiresIn into io.micronaut.security.oauth2.endpoint.token.response.TokenResponse

lightoze:
@ToniA Isn't it Keycloak-specific and not part of the standard?

ToniA:
You're right, it's not in the RFC. Anyway, access to this would be useful, as I would somehow need to be able to tell what is the expiration of the refresh token. Keycloak adds this property into the response.

Of course the value could also be calculated from the difference of 'iat' and 'exp' claim values of the refresh token.

lightoze:
@ToniA You can create a custom `OpenIdUserDetailsMapper` and it will have access to all token claims.

ToniA:
@lightoze OpenIdUserDetailsMapper has the access token claims, but not the refresh token claims. It has access to the refresh token as a string, through the refreshToken property of OpenIdTokenResponse, though.

lightoze:
@ToniA I see, I don't use Keycloak so I assumed it's a part of access token. In many OIDC implementations refresh token is opaque (just an arbitrary ID, not a JWT) so generally you don't want rely on its contents.

ToniA:
Thanks for the suggestion, the code is so much cleaner this way. So in Keycloak, the refresh token is just another JWT. So I will parse it in my custom OpenIdUserDetailsMapper and calculate the time difference between iat and exp in seconds.
```
        var refreshTokenClaims = JWTParser.parse(refreshTokenString).getJWTClaimsSet();

        var refreshTokenIat = refreshTokenClaims.getClaim(JwtClaims.ISSUED_AT) as Date;
        var refreshTokenExp = refreshTokenClaims.getClaim(JwtClaims.EXPIRATION_TIME) as Date;
        var refreshTokenExpiresIn = Duration.between(refreshTokenIat.toInstant(), refreshTokenExp.toInstant()).getSeconds();
```

PR #483, opened 2020-12-02
Use new snapshot versions

ilopmar:


sdelamo:
@ilopmar should we change the `gradle.properties`file to no longer use `BUILD-SNAPSHOT`

ilopmar:
You can manually do it if you want. It will be done automatically next time you release a new version of the module: https://github.com/micronaut-projects/github-actions/blob/master/post-release/entrypoint.sh#L51

PR #521, opened 2020-12-23
refactor: add expiration to TokenResponse

sdelamo:
This moves the logic of calculating an expiration date for a token directly into the `TokenResponse` POJO. So that it is easier to get rid of `CacheableProcessor`.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/CacheableProcessor.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/CacheableProcessor.java
@@ -25,7 +25,6 @@ import org.reactivestreams.Subscription;
 
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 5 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/CacheableProcessor.java
```diff
@@ -52,22 +51,11 @@
...
      */
     CacheableProcessor() {
-        this.transformer = null;
-    }
 
-    /**
-     *
-     * @param transformer A transformation function to be applied when an element is received.
-     */
-    CacheableProcessor(@NonNull Function<T, T> transformer) {
```
alvarosanchez, line 69: This seems a breaking change, although the class is `@Internal` and package-private. I guess it's ok

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/TokenResponseExpiration.java
```diff
@@ -43,28 +40,12 @@ public TokenResponseExpiration(TokenResponse tokenResponse) {
         setTokenType(tokenResponse.getTokenType());
     }
 
-    /**
-     * Populates Expiration date with the expires in value.
-     *
-     * @param expiresIn The lifetime in seconds of the access token.
-     */
-    @Override
-    public void setExpiresIn(Integer expiresIn) {
```
alvarosanchez, line 52: This is a breaking change as far as I can tell.

alvarosanchez (changes requested):

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/TokenResponseExpiration.java
sdelamo: I don't think it is a breaking change because there is an implementation of `setExpiresIn` in `TokenResponse` the parent class of this class. Thus, `TokenResponseExpiration` still exposes a `setExpiresIn` method.

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/CacheableProcessor.java
sdelamo: Yes, I think it is fine because it is `@Internal `

alvarosanchez (approved):

Issue #500, opened 2020-12-14
Concurrency bug in CacheableProcessor (used by DefaultClientCredentialsClient)


ritikmishra:
### Task List

- [ ] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [ ] Full description of the issue provided (see below)

### Steps to Reproduce

1. Add necessary configuration to `application.yaml` to enable the `ClientCredentialsHttpClientFilter` 
2. Make multiple, **simultaneous** API calls to an API for which the `ClientCredentialsHttpClientFilter` is going to add an auth token.
3. About 20% of the time, an `ConcurrentModificationException` will be thrown by the `CacheableProcessor` used inside the `DefaultClientCredentialsClient`. If the multiple API calls was triggered by the micronaut application trying to respond to an HTTP request, it will return a 500.

### Expected Behaviour

The requests to the API that our micronaut application depends on should never fail (as a result of the `ClientCredentialsHttpFilter`/related machinery). 

### Actual Behaviour

About 20% of the time (rough estimate, but definitely not 100%), the list `CacheableProcessor#subscriptions` is modified while `CacheableProcessor#flowData` is looping over the elements in `subscriptions`, causing a `ConcurrentModificationException` to be thrown. 

Due to its length, the stack trace is at the bottom of this issue.

### Environment Information

had this issue on 2 different computers

| **Operating System**: | **Micronaut Version:** | **JDK Version:**           |
| ----------------------------   | ------------------------------- | ------------------------ | 
| macOS mojave 10.14.6 | 2.2.1                              | openJDK 11.0.8                      |
|      Ubuntu 20.04    |       2.2.1      | openJDK 11.0.9.1  |
### Example Application

[https://github.com/ritikmishra/micronaut-client-credentials-http-filter-issue](https://github.com/ritikmishra/micronaut-client-credentials-http-filter-issue)

### Stacktrace

```
java.lang.NullPointerException: Actually not, but can't throw other exceptions due to RS
        at io.reactivex.Flowable.subscribe(Flowable.java:14927)
        at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
        at io.reactivex.Flowable.subscribe(Flowable.java:14865)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
[...]

PR #527, opened 2020-12-27
replace usage of CacheableProcessor w/ RxJava cache operator

ritikmishra:
- I had looked into using [`BehaviorProcessor`](http://reactivex.io/RxJava/javadoc/io/reactivex/processors/BehaviorProcessor.html)/`BehaviorSubject`, but it wasn't easy to know if an access token request was pending using those objects.

- This still isn't perfect. In the [example](https://github.com/ritikmishra/micronaut-client-credentials-http-filter-issue) demonstrating #500, 6 access token requests get made at once when the token is expired/not cached. Ideally, only 1 request would get made.


sdelamo:
Thanks a lot for your contribution @ritikmishra I have pull your changes into https://github.com/micronaut-projects/micronaut-security/pull/548 I think it fixes 500 completely. Let me know if you agree.

PR #569, opened 2021-02-12
Add support for Reactive security rules

Steven-M-Brown:
Adds a `ReactiveSecurityRule` and updates the `SecurityFilter` to sort and interleave reactive and normal `SecurityRule` objects in order to enable blocking work without blocking the netty event loop.

Solution based on the proposed solution offered up by @jameskleeh on Gitter.

Starting this off as a draft to get some early feedback (and tips on the reactive stuff, still learning it, might be wrong) before I start on tests and documentation.

Edit: Just wanted to note that we have been successfully running this version in our codebase now for almost two days via `@Replaces`

### DIFF
--- a/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
+++ b/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
@@ -36,6 +36,7 @@ import io.micronaut.security.rules.SecurityRule;
 import io.micronaut.security.rules.SecurityRuleResult;
 import io.micronaut.web.router.RouteMatch;
 import io.reactivex.Flowable;
+import io.reactivex.Single;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -87,6 +88,23 @@ public class SecurityFilter extends OncePerRequestHttpServerFilter {
     protected final Collection<AuthenticationFetcher> authenticationFetchers;
 
     protected final SecurityConfiguration securityConfiguration;
+
+    /**
+     * @param securityRules The list of security rules that will allow or reject the request
+     * @param authenticationFetchers List of {@link AuthenticationFetcher} beans in the context.
+     * @param securityConfiguration The security configuration
+     */
+    @Deprecated
+    public SecurityFilter(
+            Collection<SecurityRule> securityRules,
+            Collection<AuthenticationFetcher> authenticationFetchers,
+            SecurityConfiguration securityConfiguration) {
+
+        this.authenticationFetchers = authenticationFetchers;
+        this.securityConfiguration = securityConfiguration;
+        this.orderedSecurityRules = new ArrayList<>(securityRules);
+        OrderUtil.sort(this.orderedSecurityRules);
+    }
 
     /**
      * @param securityRules The list of security rules that will allow or reject the request
@@ -178,11 +196,12 @@ public class SecurityFilter extends OncePerRequestHttpServerFilter {
                                 }
                                 return Publishers.empty();
                             } else if (ordered instanceof ReactiveSecurityRule) {
-                                return Flowable.fromPublisher(
+                                return Single.fromPublisher(
                                         ((ReactiveSecurityRule) ordered).check(request, routeMatch, attributes))
                                         // Ideally should return just empty but filter the unknowns
                                         .filter((result) -> result != SecurityRuleResult.UNKNOWN)
-                                        .doAfterNext((result) -> logResult(result, method, path, ordered));
+                                        .doAfterSuccess((result) -> logResult(result, method, path, ordered))
+                                        .toFlowable();
                             } else {
                                 // Not sure how we got this object in here so log and return an empty
                                 LOG.warn("Unsupported type {}", ordered.getClass());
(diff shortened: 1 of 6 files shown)


Steven-M-Brown:
Quick update, discovered an issue with the argument binding for Authorization using this so I am digging in. I'll be sure to add tests for that case when I get the green light on the solution as it stands and I start adapting our internal tests for this to the test-suites.

Steven-M-Brown:
Okay that binding issue only impacts us because our internal version of the class we are using via `@Replaces` has a different name and the argument binders use `micronaut.once.SecurityFilter` so it was a red herring. Apologies for the confusion. Would appreciate it if someone could give this a quick look/feedback so I know if I need to change anything before adapting our tests to fit the style of this project's test suites.

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
```diff
@@ -149,32 +165,76 @@ public int getOrder() {
...
+                            if (ordered instanceof SecurityRule) {
+                                SecurityRuleResult result =
+                                        ((SecurityRule) ordered).check(request, routeMatch, attributes);
+                                if (result == SecurityRuleResult.REJECTED || result == SecurityRuleResult.ALLOWED) {
+                                    logResult(result, method, path, ordered);
+                                    return Publishers.just(result);
+                                }
+                                return Publishers.empty();
+                            } else if (ordered instanceof ReactiveSecurityRule) {
+                                return Flowable.fromPublisher(
```
jameskleeh, line 181: You should enforce only a single result by using Single here. As this is its possible to return multiple results which doesn't make sense

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
```diff
@@ -76,23 +81,34 @@
...
+     * @param securityConfiguration The security configuration
      */
     @Inject
-    public SecurityFilter(Collection<SecurityRule> securityRules,
-                          Collection<AuthenticationFetcher> authenticationFetchers,
-                          SecurityConfiguration securityConfiguration) {
-        this.securityRules = securityRules;
+    public SecurityFilter(
+            Collection<SecurityRule> securityRules,
+            Collection<ReactiveSecurityRule> reactiveSecurityRules,
```
jameskleeh, line 100: This is a breaking change. This constructor can be deprecated and a new constructor created annotated with `@Inject`

jameskleeh:
Overall looks good

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
Steven-M-Brown: I tried making this work with Single but for the life of me couldn't since it explodes on the first empty (representing unknown). I added a `firstElement()` call that should cancel after the first. Open to suggestions if you know of a way to make Single work, my reactive is kind of weak.

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
Steven-M-Brown: Updated.

Steven-M-Brown:
I converted our Java tests to Groovy and I'm not 100% sure I did everything right since this is my first set of Groovy/Spock tests I have written. Would appreciate feedback again when you get time!

Also, where should I be adding documentation for this (if at all)?

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
jameskleeh: @Steven-M-Brown If you want to support empty then you can use Maybe

security/src/test/groovy/io/micronaut/security/filters/SecurityFilterSpec.groovy
```diff
@@ -0,0 +1,121 @@
...
+class SecurityFilterSpec extends EmbeddedServerSpecification {
+
+    @Override
+    String getSpecName() {
+        'SecurityFilterSpec'
+    }
+
+    @Unroll('#description')
+    def "Filter order is correct"() {
+        given:
```
jameskleeh, line 36: In general we try to avoid tests that use mocking. This can be tested by having multiple security rules, both reactive and not that return a value based on a condition (`?ruleA=true`), for example. You can then send multiple requests with different query values to effectively control the behavior of the rules then assert the correct result is returned.

jameskleeh:
This should be documented here https://github.com/micronaut-projects/micronaut-security/blob/8393c8bfb3fed517e76e853edc408259a69e4447/src/main/docs/guide/securityRule.adoc

security/src/test/groovy/io/micronaut/security/filters/SecurityFilterSpec.groovy
jameskleeh: You could also just get a reference to the bean and set a local boolean flag

security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
Steven-M-Brown: Having trouble getting a `Maybe` from a `Publisher` short of just using `firstElement()`, suggestions?

security/src/test/groovy/io/micronaut/security/filters/SecurityFilterSpec.groovy
Steven-M-Brown: I took a stab at a no-mock version. My Groovy/Spock is still really weak as this is the first set of tests I've done this way so I'm not 100% sure if I did everything correctly.

sdelamo:
@Steven-M-Brown I am flagging this under consideration for the next mayor version of Micronaut Security. If we are changing `SecurityRules` contract, I would prefer to have a single method which returns a Publisher than two methods.

Steven-M-Brown:
Awesome, looking forward to it! Just let me know if you need anything from me on this.

jameskleeh:
Closing this in favor of https://github.com/micronaut-projects/micronaut-security/pull/747

PR #581, opened 2021-03-01
Adds JwtClaimsValidators for issuer, audience, and not-before

JasonTypesCodes:


security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/NotBeforeJwtClaimsValidator.java
```diff
@@ -0,0 +1,81 @@
...
+import io.micronaut.security.token.jwt.generator.claims.JwtClaims;
+
+/**
+ * Validate current time is not before the not-before claim of a JWT token
+ *
+ * @author Jason Schindler
+ * @since 2.4.0
+ */
+@Singleton
+@Requires(property = NotBeforeJwtClaimsValidator.NOT_BEFORE_PROP, value = StringUtils.TRUE)
```
JasonTypesCodes, line 39: I made this require the property to be true instead of enabling it by default because I presumed that enabling it by default would constitute a breaking change.

JasonTypesCodes:
I have some ideas on modifying the JwtClaims interface but didn't want to couple these validators with a breaking change.

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/IssuerJwtClaimsValidator.java
```diff
@@ -0,0 +1,78 @@
...
+    public static final String ISSUER_PROP = JwtClaimsValidator.PREFIX + ".issuer";
+
+    private static final Logger LOG = LoggerFactory.getLogger(IssuerJwtClaimsValidator.class);
+
+    private final String expectedIssuer;
+
+    public IssuerJwtClaimsValidator(
+        @Property(name = IssuerJwtClaimsValidator.ISSUER_PROP) String expectedIssuer
+    ) {
+        if (LOG.isTraceEnabled()) {
```
jameskleeh, line 46: this should be debug because this will only happen once

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/IssuerJwtClaimsValidator.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param claims JwtClaims
+     * @return True if the JWT issuer claim equals the configured value
+     */
+    @Deprecated
+    @Override
+    public boolean validate(JwtClaims claims) {
+        final String issuer = (String) claims.get(JwtClaims.ISSUER);
+
+        if (issuer == null) {
+            if (LOG.isDebugEnabled()) {
```
jameskleeh, line 63: This should be trace since it can happen for each request

sdelamo (changes requested):

sdelamo (approved):
Thanks for the PR. it looks good. @JasonTypesCodes I have done several changes. Mostly, tests refactoring. I have explained my changes through conversations.

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/IssuerJwtClaimsValidator.java
sdelamo: This is now set to debug.

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/IssuerJwtClaimsValidator.java
sdelamo: This is now set to trace.

PR #597, opened 2021-03-12
build: remove micronaut-graal annoation processor

sdelamo:


sdelamo:
I think we needed because we use `@TypeHint` in `micronaut-security`.

Issue #606, opened 2021-03-26
Security endpoints sentitive by roles


oscarcitoz:
Sensitive endpoints are exposed to all logged in users regardless of their role, that is, everyone can have access to these endpoints, which causes a security breach due to the information they handle.

It will create a pr
https://github.com/micronaut-projects/micronaut-security/pull/589

with the option to configure the users who can access them through roles (without affecting the default configuration)

PR #589, opened 2021-03-10
add logic for roles endpoints sentitive (exclude health). issue https://github.com/micronaut-projects/micronaut-security/issues/606

oscarcitoz:
This logic helps for configure roles in properties and send response UNAUTHORIZED when not apply roles (only when you configure that properties endpoints.roles), Only apply for endpoints sensitive minus health. 

This PR improves security because any logged in user can see all the specific sensitive routes

PR #619, opened 2021-04-12
docs: improve documentation about SensitiveEndpointRule replacement

sdelamo:
close: https://github.com/micronaut-projects/micronaut-security/issues/606
close: https://github.com/micronaut-projects/micronaut-security/pull/589

@oscarcitoz this PR refactors `SensitiveEndpointRule` to ease replacement. Moreover, it documents how to restrict management endpoints for a particular role.

### DIFF
--- a/security/src/main/java/io/micronaut/security/rules/SensitiveEndpointRule.java
+++ b/security/src/main/java/io/micronaut/security/rules/SensitiveEndpointRule.java
@@ -15,15 +15,25 @@
  */
 package io.micronaut.security.rules;
 
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.inject.ExecutableMethod;
 import io.micronaut.management.endpoint.EndpointSensitivityProcessor;
+import io.micronaut.management.endpoint.beans.BeansEndpoint;
+import io.micronaut.management.endpoint.env.EnvironmentEndpoint;
+import io.micronaut.management.endpoint.health.HealthEndpoint;
+import io.micronaut.management.endpoint.info.InfoEndpoint;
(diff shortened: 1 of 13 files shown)


oscarcitoz (approved):
approved, it hurts that it could not be done with my PR because I did it so that it was not necessary to create a new class to specify the role but from configurations (from applicaiton.yml). anyway thank you very much

security/src/main/java/io/micronaut/security/rules/SensitiveEndpointRule.java
```diff
@@ -57,35 +76,122 @@
...
+                                       @NonNull ExecutableMethod<?, ?> method) {
+        Boolean sensitive = endpointMethods.get(method);
+        if (sensitive) {
+            if (claims == null) {
+                return checkSensitiveAnonymous(request, method);
+            }
+            return checkSensitiveAuthenticated(request, claims, method);
+        }
+        return checkNotSensitive(request, claims, method);
+    }
```
ilopmar, line 113: I'm not sure if this is the best way to encapsulate everything calling the threes protected methods from here. All three methods are the same in the sense that they log a message and return the appropriate `SecurityRuleResult` but they don't check any condition. Everything is decided in this method.
Did you create those three methods in other for users to be able to override them without override this one?

sdelamo: Yes, three protected methods makes easier the replacement. See example, in the tests which I show in the docs.

ilopmar: :+1:

ilopmar (approved):

sdelamo:
thanks @oscarcitoz for pointing us to improving this part of the security module. Really valuable feedback.

PR #633, opened 2021-04-27
Make public access to getCookies methods in CookieLoginHandler class.

donbeave:
I'm using this module with GraphQL login implementation where I need to call some classes directly, and instead of copy-paste the whole method to generate cookies from `UserDetails` object it will be better to open access to these methods. This will help to reuse these methods after upgrade to the new version.

This is an example of how I use this code in GraphQL endpoints: https://github.com/expatiat/dumper/blob/main/dumper-api/src/main/java/com/zhokhov/dumper/api/graphql/mutation/UserLoginServerMutation.java

This method is mostly copy-pasted from `JwtCookieLoginHandler` class:
`private Optional<Cookie> accessTokenCookie(UserDetails userDetails, HttpRequest<?> request)`


sdelamo:
@donbeave I am ok making those methods public but we need to add javadoc to them.

donbeave:
@sdelamo done, please check https://github.com/micronaut-projects/micronaut-security/pull/654

PR #683, opened 2021-06-10
Log on missing JWT configs in validation

sascha-frinken:
Add logging when validating encrypted or signed JWTs and there is no config.


sdelamo (approved):
Thanks @sascha-frinken

PR #721, opened 2021-07-13
Fix: JWT Refresh Token Cookie not getting cleared (#718)

### LINKED ISSUE #718: JWT refresh token cookie will not be cleared on logout
### Steps to Reproduce

1. Configure your app to use JWT cookie auth
2. Login with the default `JWTCookieLoginHandler` (`/login`)
3. Logout with the default `JWTCookieClearerLogoutHandler` (`/logout`)

### Expected Behaviour

The `JWT_REFRESH_TOKEN` should be cleared.

### Actual Behaviour

The `JWT_REFRESH_TOKEN` is not cleared. I assume this is the case because in `JWTCookieClearerLogoutHandler#clearCookie()` the cookie will only be cleared if a cookie with the same name is present. However, on `/logout` the browser wont send cookies with `/oauth/access_token` domain set. Here's some screenshots:

Screenshot 2021-07-07 at 15 07 16

Screenshot 2021-07-07 at 15 07 38

Screenshot 2021-07-07 at 15 09 14

I reckon a fix is easy as the logout handler should set a "cleared" cookie anyway, even if the cookie is not found. Is this correct?

### Environment Information

- **Operating System**: doesn't matter
- **Micronaut Version:** 3.0.0-M2 with netty server
- **JDK Version:** 16

logineimer:
Fixes #718 by clearing the Token and Refresh Token Cookies anyway.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/cookie/JwtCookieClearerLogoutHandler.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/cookie/JwtCookieClearerLogoutHandler.java
@@ -27,7 +27,6 @@ import io.micronaut.security.handlers.LogoutHandler;
 import jakarta.inject.Singleton;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.Optional;
 
 /**
  * Clears the cookie configured via {@link CookieLoginHandler}.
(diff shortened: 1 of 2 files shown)


graemerocher:
Needs a test

logineimer:
> Needs a test

Sure. Will contributors do it or did I miss something?

logineimer:
> Edit: It took a while to understand. I wrote a spec for the fix. Hope it's satisfactory. 🤣

I don't know if you got a notification when I edited my comment above. It's been ten days now and nothing happened. 🥲🤔 Spoke with a security expert at my company today about this bug. He says it's a serious one because the user isn't able to really logout without the fix. So I hope the fix will be merged soon. 😇

Issue #308, opened 2020-07-11
Authentication attributes should be convertable


jameskleeh:
It would be nice to do `getAttributes.get("id", Long.class)` for example

p1729:
@jameskleeh I would like to contribute to this. Are you looking for something like a default method in `io.micronaut.security.authentication.Authentication` interface
```
default <T> Optional<T> getAttribute(String name, Class<T> clazz) {
        return Optional.ofNullable(getAttributes().get(name))
                .filter(clazz::isInstance)
                .map(clazz::cast);
}
```

ojebari-ma:
Can you please confirm that this is still an issue in the latest version of Micronaut?

sdelamo:
> [@jameskleeh](https://github.com/jameskleeh) I would like to contribute to this. Are you looking for something like a default method in `io.micronaut.security.authentication.Authentication` interface
> 
> ```
> default <T> Optional<T> getAttribute(String name, Class<T> clazz) {
>         return Optional.ofNullable(getAttributes().get(name))
>                 .filter(clazz::isInstance)
>                 .map(clazz::cast);
> }
> ```

I think @jameskleeh refers to implementing [ConvertibleValues.html](https://docs.micronaut.io/4.8.11/api/io/micronaut/core/convert/value/ConvertibleValues.html)

PR #724, opened 2021-07-14
Feat: Support for explicitly typed attributes in Authentication (#308)

logineimer:
Adds support for explicitly typed attributes in Authentication like @p1729 suggested.

### DIFF
--- a/security/src/main/java/io/micronaut/security/authentication/Authentication.java
+++ b/security/src/main/java/io/micronaut/security/authentication/Authentication.java
@@ -19,6 +19,7 @@ import io.micronaut.core.annotation.NonNull;
 import java.io.Serializable;
 import java.security.Principal;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * Represents the state of an authentication.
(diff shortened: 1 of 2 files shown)


security/src/main/java/io/micronaut/security/authentication/Authentication.java
```diff
@@ -40,4 +41,18 @@
...
+
+    /**
+     * Retrieves an explicitly typed attribute value.
+     *
+     * @param name the name of the attribute
+     * @param clazz the expected class of the attribute
+     * @param <T> the expected type of the attribute
+     * @return the typed attribute or null if it doesn't exist or the type is not correct
+     */
+    default <T> Optional<T> getAttribute(String name, @NonNull Class<T> clazz) {
```
sdelamo, line 53: Could you annotate the return type with `@NonNull`

Could you annotate the parameter `name` with `@NonNull`

sdelamo (changes requested):

sdelamo:
Thanks for the contribution. 

However, I am not sure if we should add this method. I think the recommended way do conversion is to use the [ConversionService](https://docs.micronaut.io/latest/api/index.html).

sdelamo:
@jameskleeh thoughts ?

jameskleeh:
The issue with this is that authentication objects are created outside the context and we would need a conversion service instance to do the conversion. There is ConversionService.SHARED we could use but it wouldn't include any converters added via beans, so less than ideal.

sdelamo:
I am going to close it. My recommendation is for the user to inject `ConversionService` in their singleton or bean where they are querying the authentication. I don't want to add a method to `Authentication`which does not represent best practice.

PR #735, opened 2021-07-30
Change cookie secure configuration to no longer have a default

jameskleeh:
The new default value is request.isSecure()


sdelamo:
I have documented the breaking change and remove the changes I had done in the test.

PR #742, opened 2021-08-04
Add getRoles method to Authentication and dont require TokenConfiguration to construct it

jameskleeh:
- This breaks the contract of SecurityRule to take in the Authentication object instead of a Map, which is better imo
- The authentication now has 2 versions, one for client and one for server. The client one can be used for JSON deserialization and is chosen automatically due to our jackson module.
- I had to devise a way for the client authentication to provide a useful implementation of getRoles. I did this by setting a `rolesKey` in the claims on the server, which the client authentication reads and returns the data found at that key. I don't think we are really responsible here for implementing this on the client side and would be ok with throwing an exception or just returning an empty list instead. I'm not super keen on the implementation so lmk. Another alternative is to move the `getRoles()` to a `ServerAuthentication` interface, but I don't think it makes sense to do that.


jameskleeh:
Not sure why it didn't run the tests, but they passed locally for me

sdelamo:
I like that you have renamed the `AuthenticationResponse.build` to `success` and that you have created `failure` methods. I have created `exception` methods as well.

sdelamo:
> This breaks the contract of SecurityRule to take in the Authentication object instead of a Map, which is better imo

Yes, it is better. I have documented the change in the https://github.com/micronaut-projects/micronaut-security/pull/742/commits/cd4ca0fb3dfd6ab43528fa9645ff9e03ed6e9219.

sdelamo:
I have done this change https://github.com/micronaut-projects/micronaut-security/pull/742/commits/ba43a18ee9b50c433ce3c94240a5366a4b0326ac so that we don't generate the extra claim `rolesKey` in the scenarios where the user has changed the default value. 

I have added two tests `RolesNameSpec` and `RolesNameCustomSpec` which I think illustrate two scenarios:

-  validating a JWT generated by someone else with a custom role claim. 
- validating a JWT generated by the Micronaut app with a custom role claim.

sdelamo:
I've created a PR to delete ClientAuthentication https://github.com/micronaut-projects/micronaut-security/pull/743 Do we need that?

sdelamo (commented):
The API looks better without `TokenConfiguration` but if possible I think we should remove [ClientAuthentication](https://github.com/micronaut-projects/micronaut-security/pull/743)

ilopmar:
@jameskleeh The tests don't run because you are sending a PR to another branch and they only run for `master` and `'[1-9]+.[0-9]+.x'` branches. See https://github.com/micronaut-projects/micronaut-security/blob/master/.github/workflows/gradle.yml#L14-L15

sdelamo:
@jameskleeh  I added a method to `ServerAuthentication` annotated with `@JsonValue` to generate a Json representation as the one `ClientAuthentication` expects. see: 

https://github.com/micronaut-projects/micronaut-security/pull/742/commits/61696e9d531c42c8f62df068345543d89e92f222

sdelamo (approved):

PR #743, opened 2021-08-05
Delete ClientAuthentication

sdelamo:
In which scenario do we need the JSON serialisation capability of Authentication? Could we just remove it?


sdelamo:
I discussed this with @jameskleeh and we are going to keep it.

Issue #689, opened 2021-06-17
Request body is not present in `SecurityRule`


tcrespog:
### Task List

- [x] Steps to reproduce provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Download the [example application](https://github.com/tcrespog/mn-security-rule-body).
2. Run the `PetControllerTest`, which performs a request to an endpoint secured with a `CustomSecurityRule`.
3. The `CustomSecurityRule` throws an exception because it expects the request body to be present (but it's not).

### Expected Behaviour

The request body should be populated by the time the security rule is processed.

### Actual Behaviour

The request body is not populated by the time the security rule is processed.

I have a custom [`SecurityRule`](https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/rules/SecurityRule.java) which reads data from the request body. However, when upgrading the app to Micronaut 2.5.6, I've realized that the body is not present in the `request` by the time the security rule is executed.
Apparently, this is happening since Micronaut 2.5.0.

### Environment Information

- **Micronaut Version:** 2.5.6
- **JDK Version:** 11

### Example Application

- https://github.com/tcrespog/mn-security-rule-body

tcrespog:
I found this PR which explains the change of behavior: https://github.com/micronaut-projects/micronaut-core/pull/5261.

Now I understand that the absence of body in filters is intentional, therefore it won't be present in security rules as they are executed in the [`SecurityFilter`](https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java#L144).

Alternatively to security rules, [AOP Around Advice](https://docs.micronaut.io/2.5.7/guide/#aroundAdvice) could be a way of handling the case described in this issue as suggested by this comment: https://github.com/micronaut-projects/micronaut-security/discussions/649#discussioncomment-966274.
Am I right?

Thanks.

sdelamo:
Yes, the absence of body in filters is intentional.

sdelamo:
> Alternatively to security rules, AOP Around Advice could be a way of handling the case described in this issue as suggested by this comment: #649 (comment).

Yes, I am going to write a documentation snippet. We should have clarified this.

sdelamo:
I am closing this issue. I have added a documentation snippet which explains that you cannot read HTTP Body from a `SecurityRule`

PR #756, opened 2021-08-13
doc: example evaluating HTTP Request Body for a SecurityRule like decision

sdelamo:
close https://github.com/micronaut-projects/micronaut-security/issues/689

### DIFF
--- a/src/main/docs/guide/securityRule.adoc
+++ b/src/main/docs/guide/securityRule.adoc
@@ -21,6 +21,5 @@ Security rules return a api:security.rules.SecurityRuleResult[]. See the followi
 
 IMPORTANT: If all security rules return `UNKNOWN`, the request will be rejected!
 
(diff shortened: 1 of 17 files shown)


sdelamo:
@jameskleeh could you please review this?

jameskleeh:
I'm not really sure we should be documenting this. From the use cases I've heard, the requirement was to verify the body matches a given signature. For that type of use case using a Netty handler would be far more reliable. `request.getBody(Map)` infers that the body can be converted to a Map, which isn't always the case. Can you explain the use case you're trying to document?

sdelamo:
After discussing this with @jameskleeh we are going to close this PR. He is going to document how to use a Netty Handler for this use case.

kanak-kabara:
@jameskleeh Has this been documented anywhere, apologies if it has.

Issue #751, opened 2021-08-10
Support for including an audience parameter when requesting OAuth token


sturlapati-adaptive:
### Feature description

Auth0 expects to receive the audience parameter when requesting a token from them. This applies to following flows: AuthorizationCode, ClientCredentials, Password.

Please provide a way to specify the audience value in the oauth configuration when requesting a token.

NOTE: I already wrote the code and I can do pull request. I need some guidance on the expected tests.

thanks

sturlapati-adaptive:
Can you please give me permission to push changes to a branch so i can create a PR?

sdelamo:
@sturlapati-adaptive you will need to fork the repository and create a pull request from your fork.

sdelamo:
@sturlapati-adaptive I just created a Micronaut application and used Auth0 and I did not need to supply the audience. 

I have written a guide: https://github.com/micronaut-projects/micronaut-guides/pull/500 

Can you point me what I am missing ?

imre-kerr:
I'm not sure why it's not required in that case, but at least for Client Credentials flow the docs say it's required: https://auth0.com/docs/authorization/flows/call-your-api-using-the-client-credentials-flow#request-tokens

And the response (HTTP 403) when it's missing seems to confirm it:
```json
{"error":"access_denied","error_description":"Non-global clients are not allowed access to APIv1"}
```

jameskleeh:
Because "audience" is not part of the standard I'm not sure we should support it explicitly, but perhaps with a map of configurable claims:

```
client-credentials:
    additional-claims:
        audience: XYZ
```

frehov:
I've had a look at the backing code for the client credentials.
I've especially had a look at the `ClientCredentialsTokenRequestContext` which seems like it would be the most elegant place to inject those extra request parameters.

By hooking up the `ClientCredentialsTokenRequestContext` as the place to inject the extra parameters it's less vulnerable to someone overwriting parts of the secure grant map, unless they're really trying to.

This is the following change I'm proposing, and I can open a pull request with most of the relevant code if that's of interest.
```java
   private final Map<String, String> additionalRequestParams;

    @Override
    public Map<String, String> getGrant() {
        Map<String, String> grantMap = grant.toMap();
        grantMap.putAll(additionalRequestParams);
        return grantMap;
    }
```

sturlapati-adaptive:
[...]

PR #766, opened 2021-08-26
add audience configuration property and populate the appropriate Gran…

sturlapati-adaptive:
Fixes micronaut-projects/micronaut-security#751 add audience configuration property and populate the appropriate grant objects with the value from the config.


sdelamo:
I did not need to supply an audience here: https://github.com/micronaut-projects/micronaut-guides/pull/500

sdelamo:
Thanks for the PR. However, We are going to solve this via https://github.com/micronaut-projects/micronaut-security/pull/776

PR #779, opened 2021-09-14
feat: SecurityFilter toggeable and configurable pattern

### LINKED ISSUE #550: Cannot disable endpoint security and enable HTTP client credentials filter
I am building a native image lamba with Micronaut that is called via AWS API Gateway (APIG). Auth is handled by APIG so is not required on the apps controller. The app has a client that uses OAUTH2 client credentials to call a secured external API. I cannot find a way to disable endpoint security but enable the HTTP client filter for client credentials `ClientCredentialsHttpClientFilter`.

### Expected Behaviour

The following config should disable the security requirement for controllers **and** enable OAUTH2 client credentials for the  HTTP client.

```yaml
micronaut:
  http:
    services:
      myClient:
        url: 'api.abc.com'
  security:
    enabled: false
    oauth2:
      enabled: true
      clients:
        abc:
          enabled: true
          client-id: myClientId
          client-secret: myClientSecret
          scopes:
            - myScope
          client-credentials:
            enabled: true
            service-id-regex: 'myClient'
          token:
            url: 'https://login.microsoftonline.com/myAccountID/oauth2/v2.0/token'
            auth-method: client_secret_basic
          grant-type: client_credentials
```

### Actual Behaviour

No `ClientCredentialsHttpClientFilter` bean is created and the `abc` client does not login to https://login.microsoftonline.com/myAccountID/oauth2/v2.0/token to get the `Bearer` token or add it to the client request.

### Environment Information

- **Operating System**: linux
- **Micronaut Version:** 2.3.0
- **JDK Version:** 11

At the moment I am leaving security enabled and adding `@Secured(SecurityRule.IS_ANONYMOUS)` to the controller as a workaround.

```yaml
  security:
    endpoints:
      login:
        enabled: false
```

sdelamo:
This is useful for people who want to use Micronaut security features (token validation, client credentials) but they don't need route validation. 

Closes #550

### DIFF
--- a/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
+++ b/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java
@@ -15,10 +15,11 @@
  */
 package io.micronaut.security.filters;
 
-
 import io.micronaut.context.annotation.Replaces;
+import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.order.Ordered;
+import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.HttpAttributes;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpStatus;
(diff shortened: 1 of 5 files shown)


sdelamo:
@jameskleeh what do you think?

Issue #771, opened 2021-09-02
JWKS not been cached


sdelamo:
### Discussed in https://github.com/micronaut-projects/micronaut-security/discussions/765

<div type='discussions-op-text'>

<sup>Originally posted by **enzomotta** August 24, 2021</sup>
Hi there,

I was using only AWS Cognito to authenticate my users and everything was perfect til I had to support Google and Apple native logins. So, to validate those jwts so I added the respective JWKS on my application.yml and I'm validating issuer and audience on a custom JwtAuthenticationFactory implementation. After that I notice that every request I received had Micronaut downloading the JWKS jsons from Google and Apple. Wasn't it supposed to cache those calls? It appears to me that the cognito JWKS is been cached cause its not been downloaded all the time.

My application.yml looks something like this:
```        
signatures:
  jwks:
    awscognito:
      url: 'https://cognito-idp.us-east-1.amazonaws.com/${integrations.cognito.userPoolId}/.well-known/jwks.json'
    google:
      url: 'https://www.googleapis.com/oauth2/v3/certs'
    apple:
      url: 'https://appleid.apple.com/auth/keys'
```

Am I missing something?

Thanks in advance,

Enzo.</div>

Steven-M-Brown:
We see this too; it seems to consider each jwk source independently after sorting.

This results in the following behaviors (using your yml as an example and the assumption that is the sort order):
1. Get jwt signed by awscognito:
    * checks awscognito, passes, continues.
2. Get jwt signed by google:
    * checks awscognito, fails, refreshes awscognito, still fails
    * checks google, passes, continues
3. Get jwt signed by apple:
    * checks awscognito, fails, refreshes awscognito, still fails
    * checks google, fails, refreshes google, still fails
    * checks apple, passes, continues
4. Attacker sends jwt with key not in any:
    * checks awscognito, fails, refreshes awscognito, still fails
    * checks google, fails, refreshes google, still fails
    * checks apple, fails, refreshes apple, still fails

In reality it should check all, then refresh all if not in any (to handle rotations), then check all again. There should also be some sort of rate limiting on the refresh since we don't want situation 4 to result in fetching every jwk for each request sent.

sdelamo:
@Steven-M-Brown I have a PR which I think helps. Can you check it out and let me know what you think? 

https://github.com/micronaut-projects/micronaut-security/pull/823

Steven-M-Brown:
[...]

PR #823, opened 2021-10-26
fix: improve JWKS caching

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-security/issues/771

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
@@ -37,6 +37,7 @@ import java.net.URL;
 import java.text.ParseException;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
@@ -61,7 +62,7 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
     private static final Logger LOG = LoggerFactory.getLogger(JwksSignature.class);
     private final JwkValidator jwkValidator;
     private final JwksSignatureConfiguration jwksSignatureConfiguration;
-    private volatile Date jwkSetCachedAt;
+    private volatile Instant jwkSetCachedAt;
     private volatile JWKSet jwkSet;
 
     /**
@@ -80,7 +81,7 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
      * @param url The JWK url
      * @param keyType The JWK key type
      * @param jwkValidator JWK Validator to be used.
-     * @deprecated Use {@link JwksSignature(JwksSignatureConfiguration,JwkValidator)} instead.
+     * @deprecated Use {@link #JwksSignature(JwksSignatureConfiguration, JwkValidator)} instead.
      */
     @Deprecated
     public JwksSignature(String url,
@@ -118,7 +119,7 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
                 if (jwkSet == null) {
                     jwkSet = loadJwkSet(this.jwksSignatureConfiguration.getUrl());
                     this.jwkSet = jwkSet;
-                    this.jwkSetCachedAt = Date.from(Instant.now().plus(this.jwksSignatureConfiguration.getCacheExpiration(), ChronoUnit.SECONDS));
+                    this.jwkSetCachedAt = Instant.now().plus(this.jwksSignatureConfiguration.getCacheExpiration(), ChronoUnit.SECONDS);
                 }
             }
         }
@@ -130,28 +131,41 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
     }
 
     @Override
+    public boolean isExpired() {
+        Instant cachedAt = jwkSetCachedAt;
+        return cachedAt != null && Instant.now().isBefore(cachedAt);
+    }
+
+    @Override
+    public void clear() {
+        jwkSet = null;
+        jwkSetCachedAt = null;
+    }
+
+    @Override
+    public boolean isPresent() {
+        return jwkSet != null;
+    }
+
+    @Override
     @NonNull
-    public synchronized Optional<Boolean> isJwksCacheExpired() {
-        Date cachedAt = this.jwkSetCachedAt;
-        return isExpired(cachedAt);
-    }
-
-    @Override
-    public synchronized void clearJwksCache() {
-        this.jwkSet = null;
-        this.jwkSetCachedAt = null;
-    }
-
-    @Override
-    public synchronized boolean isJwksCachePresent() {
-        return this.jwkSet != null;
-    }
-
-    @Override
-    @NonNull
-    public synchronized Optional<List<String>> getJwksKeyIDs() {
+    public Optional<List<String>> getKeyIds() {
         JWKSet jwkSet = this.jwkSet;
-        return parseJwksKeyIDs(jwkSet);
+        if (jwkSet != null) {
+            List<String> keyIds = new ArrayList<>(jwkSet.getKeys().size());
+            for (JWK key : jwkSet.getKeys()) {
+                String keyId = key.getKeyID();
+                if (keyId != null) {
+                    keyIds.add(keyId);
+                }
+            }
+            if (keyIds.isEmpty()) {
+                return Optional.empty();
+            } else {
+                return Optional.of(keyIds);
+            }
+        }
+        return Optional.empty();
     }
 
     /**
@@ -259,7 +273,7 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
                 LOG.debug("JWT Key ID: {}", keyId);
             }
             if (LOG.isDebugEnabled()) {
-                LOG.debug("JWK Set Key IDs: {}", String.join(",", getJwksKeyIDs().orElse(Collections.emptyList())));
+                LOG.debug("JWK Set Key IDs: {}", String.join(",", getKeyIds().orElse(Collections.emptyList())));
             }
             if (keyId != null) {
                 builder = builder.keyID(keyId);
@@ -268,20 +282,6 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
             matches = new JWKSelector(builder.build()).select(jwkSet);
         }
         return matches;
-    }
-
-    @NonNull
-    private static Optional<List<String>> parseJwksKeyIDs(@Nullable JWKSet jwkSet) {
-        return jwkSet == null ? Optional.empty() : Optional.of(jwkSet.getKeys()
-                .stream()
-                .map(JWK::getKeyID)
-                .filter(Objects::nonNull)
-                .collect(Collectors.toList()));
-    }
-
-    @NonNull
-    private static Optional<Boolean> isExpired(@Nullable Date cachedAt) {
-        return (cachedAt == null) ? Optional.empty() : Optional.of(!new Date().before(cachedAt));
     }
 
     /**
@@ -303,7 +303,7 @@ public class JwksSignature implements JwksCache, SignatureConfiguration {
     public int getRefreshJwksAttempts() {
         return DEFAULT_REFRESH_JWKS_ATTEMPTS;
     }
-    
+
     /**
      *
      * @return A JSON Web Key Validator.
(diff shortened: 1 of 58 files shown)


Steven-M-Brown (commented):
I don't think this actually fixes the issues in #771 as it still considers each set of jwk's independently due to the `@EachBean(JwksSignatureConfiguration.class)`.

I think this might actually be more broken now in as it won't ever refresh on a miss seemingly.

Consider this:
Key doesn't match anything in the jwk set: will return false out of verify.
Key does match but fails verification (not sure when this would actually even happen): triggers refresh by nuking the set and recurses.

The key not matching anything is the one that needs to trigger the refresh to handle a singing key rotation (signed by a new key id, causes a no-match) but it needs to consider all available JWKSets across all available `JwksSignature` before doing the refresh (original issue when more than one source for jwk's is used).

On an unrelated note I think your double-checked locking usage in `getJWKSet` is incorrect (missing volatile on `jwkSet`).

Steven-M-Brown:
If it would be helpful I can try and clean up our bandaid fix and put up a PR just so you can get an idea of what amounted to a bit of a hack to solve it.

sdelamo:
> B) Key does match but fails verification (not sure when this would actually even happen): triggers refresh by nuking the set and recurses.

I think that could happen when an authorization server rotates key without changing the key id. That should not be done by an authorization server but it is an scenario.

Steven-M-Brown:
> I think that could happen when an authorization server rotates key without changing the key id. That should not be done by an authorization server but it is an scenario.
Okay that makes sense I guess (despite feeling really wrong to do I'm sure it happens).

sdelamo:
I have added a test and code to handle scenario C. 

https://github.com/micronaut-projects/micronaut-security/pull/823/commits/07e55a694425a3224b1c691c6a40b8cf86ec2eab

Let me know what you think @Steven-M-Brown

sdelamo:
I have added a test for no `kid`. https://github.com/micronaut-projects/micronaut-security/pull/823/commits/96fb057a4e2cdd677a8543f768a88168c9658d00 If I think that covers A. Not sure what else we can do about it. What do you think @Steven-M-Brown ?

sdelamo:
someone could attack the server by sending JWT signed multiple times. Such attack may cause the server to hit the remote JWKS constantly. To avoid that I have added a configuration property set to 60 seconds. 

`'micronaut.security.token.jwt.signatures.jwks.*.cache-expiration`. If the token verification fails, The JWKS are not refreshed from remote server until the cache has expired.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -152,6 +184,19 @@ public boolean verify(SignedJWT jwt) throws JOSEException {
...
+        }
+    }
+
+    @Override
+    public boolean isJwksCachePresent() {
+        return this.jwkSet != null;
+    }
+
     /**
      * Instantiates a JWKSet for a give url.
```
Steven-M-Brown, line 201:
```suggestion
     * Instantiates a JWKSet for a given url.
```

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -182,30 +226,60 @@ protected JWKSet loadJwkSet(String url) {
...
      */
+    @Deprecated
     protected List<JWK> matches(SignedJWT jwt, @Nullable JWKSet jwkSet, int refreshKeysAttempts) {
+        return matches(jwt, jwkSet);
+    }
+
+    /**
+     * Calculates a list of JWK matches for a JWT.
+     *
+     * Since the JWTSet is cached it attempts to refresh it (by calling its self recursive)
```
Steven-M-Brown, line 240:
```suggestion
     * Since the JWKSet is cached it attempts to refresh it (by calling itself recursively)
```

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -47,17 +53,16 @@
...
+    @Deprecated
     public static final int DEFAULT_REFRESH_JWKS_ATTEMPTS = 1;
 
     private static final Logger LOG = LoggerFactory.getLogger(JwksSignature.class);
-
     private final JwkValidator jwkValidator;
-
+    private final JwksSignatureConfiguration jwksSignatureConfiguration;
+    private Date jwkSetCachedAt;
     private JWKSet jwkSet;
```
Steven-M-Brown, line 65: I still think this needs a `volatile` to be correct.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
Steven-M-Brown: Also it doesn't seem to call itself recursively anymore, is this just an out of date comment?

Steven-M-Brown (commented):
I'm finding the diff a tad hard to follow so I may have to pull down the branch and read it that way. There are a few minor grammar/javadoc mistakes and a couple possible race conditions plus a missing volatile but it looks okay as far as I can tell at this point.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
sdelamo: I have remove the line from the java doc. Please, note that I have deprecated the method.

sdelamo: thanks, I removed it.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -104,6 +129,13 @@ public JwksSignature(String url,
         return getJWKSet().map(JWKSet::getKeys).orElse(Collections.emptyList());
     }
 
+    @Override
+    @NonNull
+    public Optional<Boolean> isJwksCacheExpired() {
+        return (jwkSetCachedAt == null) ? Optional.empty() :
+                Optional.of(!new Date().before(jwkSetCachedAt));
```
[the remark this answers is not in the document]
sdelamo: I've added  `sychronized` to the method and also extracted a variable.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
sdelamo: I've added `volatile` to both.

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -131,9 +131,27 @@ public Integer getCacheExpiration() {
...
+    @Override
+    public synchronized boolean isJwksCachePresent() {
+        return this.jwkSet != null;
+    }
+
+    @Override
+    @NonNull
+    public synchronized Optional<List<String>> getJwksKeyIDs() {
+        JWKSet jwkSet = this.jwkSet;
+        return parseJwksKeyIDs(jwkSet);
```
Steven-M-Brown, line 154: So while this does fix it I don't think it was the right fix as it adds *a ton* of potential contention between requests. Note that a few of these are used by `comparator` from `validate` which is in the hot path of requests.

Rather than a sledgehammer (synchronized) I would recommend a scalpel (store it in a local variable so it doesn't check the null then NPE a couple instructions later). In this case it is actually probably okay if it compares against slightly out of date data based on the intended usages in order to not block other requests.

sdelamo:
@jameskleeh can you review this?

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -104,6 +129,31 @@ public JwksSignature(String url,
         return getJWKSet().map(JWKSet::getKeys).orElse(Collections.emptyList());
     }
 
+    @Override
+    @NonNull
+    public synchronized Optional<Boolean> isJwksCacheExpired() {
```
jameskleeh, line 134: I don't think this needs to be synchronized

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -104,6 +129,31 @@ public JwksSignature(String url,
...
+        Date cachedAt = this.jwkSetCachedAt;
+        return isExpired(cachedAt);
+    }
+
+    @Override
+    public synchronized void clearJwksCache() {
```
jameskleeh, line 140: I don't think this needs to be synchronized

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -104,6 +129,31 @@ public JwksSignature(String url,
...
+        this.jwkSet = null;
+        this.jwkSetCachedAt = null;
+    }
+
+    @Override
+    public synchronized boolean isJwksCachePresent() {
```
jameskleeh, line 146: I don't think this needs to be synchronized

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
jameskleeh: Agreed I don't see the need for synchronization here

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -169,43 +219,71 @@ protected JWKSet loadJwkSet(String url) {
...
+    private static Optional<List<String>> parseJwksKeyIDs(@Nullable JWKSet jwkSet) {
+        return jwkSet == null ? Optional.empty() : Optional.of(jwkSet.getKeys()
+                .stream()
+                .map(JWK::getKeyID)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList()));
+    }
+
+    @NonNull
+    private static Optional<Boolean> isExpired(@Nullable Date cachedAt) {
```
jameskleeh, line 283: Should probably use Instant instead of Date

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtValidator.java
```diff
@@ -191,38 +193,87 @@ private boolean hasAtLeastTwoDots(String token) {
...
+            Optional<JWT> optionalJWT = validate(jwt, config);
+            if (optionalJWT.isPresent()) {
+                return optionalJWT;
+            }
+        }
         return Optional.empty();
     }
 
-    private static Comparator<SignatureConfiguration> comparator(JWSAlgorithm algorithm) {
+    private static int compareMatchingKids(SignatureConfiguration sig,
```
jameskleeh, line 251: Is this supposed to mean `Key IDs`? If so I would have that has the name because what this method does isn't clear by the name

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtValidator.java
```diff
@@ -191,38 +193,87 @@ private boolean hasAtLeastTwoDots(String token) {
...
-            } catch (final JOSEException e) {
-                if (LOG.isDebugEnabled()) {
-                    LOG.debug("Verification failed with signature configuration: {}, passing to the next one", config);
+        Optional<JWT> optionalJWT = validate(jwt, sortedConfigs);
+        if (optionalJWT.isPresent()) {
+            return optionalJWT;
+        }
+
+        // If any of the signature configurations is a JwksCache, evict the cache and attempt to verify again
+        for (SignatureConfiguration c : sortedConfigs.stream()
```
jameskleeh, line 205: Don't use a stream here. You can just check for instanceof in the loop

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
jameskleeh: I think I take that back because we need synchronization because we have separate operations, check for null, get key set

PR #829, opened 2021-11-01
refactor: add @NonNull to X509Authentication

sdelamo:


security/src/main/java/io/micronaut/security/x509/X509Authentication.java
```diff
@@ -27,16 +28,18 @@
  */
 public class X509Authentication extends ServerAuthentication {
 
+    @NonNull
```
burtbeckwith, line 31: is this needed (and checked)? I would expect the annotation on the constructor arg and getter to be sufficient

burtbeckwith (approved):
thanks

security/src/main/java/io/micronaut/security/x509/X509Authentication.java
sdelamo: I typically annotate the field and the generate constructor and getter.

PR #853, opened 2021-11-23
Change method names. Remove synchronization. Remove use of streams

jameskleeh:


security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -130,28 +131,41 @@ public Integer getCacheExpiration() {
     }
 
     @Override
-    @NonNull
-    public synchronized Optional<Boolean> isJwksCacheExpired() {
-        Date cachedAt = this.jwkSetCachedAt;
-        return isExpired(cachedAt);
+    public boolean isExpired() {
+        Instant cachedAt = jwkSetCachedAt;
```
jameskleeh, line 135: This shouldn't need synchronization because once we store the value it doesn't matter if its overwritten before this method completes

PR #854, opened 2021-11-23
update build

sdelamo:
This updates the build to be as close as possible to the template project.

### DIFF
--- a/security-jwt/build.gradle
+++ b/security-jwt/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
+
 ext {
     nimbusJoseJwtVersion = '9.15.2'
     bouncyCastleVersion = '1.69'
@@ -32,4 +36,4 @@ test {
     testLogging.exceptionFormat = 'full'
 }
 
-//compileTestGroovy.groovyOptions.forkOptions.jvmArgs = ['-Xdebug', '-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005']
+group "io.micronaut.security"
--- a/security-jwt/build.gradle
+++ b/security-jwt/build.gradle
@@ -8,6 +8,7 @@ ext {
 }
 
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
     api "io.micronaut:micronaut-http"
     api "io.micronaut:micronaut-http-server"
     api project(":security")
--- a/security-ldap/build.gradle
+++ b/security-ldap/build.gradle
@@ -3,6 +3,8 @@ plugins {
 }
 
 dependencies {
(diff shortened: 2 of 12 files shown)


security-jwt/build.gradle
```diff
@@ -1,9 +1,14 @@
...
+    id "io.micronaut.build.internal.module"
+}
+
 ext {
     nimbusJoseJwtVersion = '9.15.2'
     bouncyCastleVersion = '1.69'
 }
 
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
```
sdelamo, line 11: I think that this should not be necessary. we were applying this in the root build to every submodule. Thus, I have keep it to avoid doing a breaking change accidentally.

security-ldap/build.gradle
```diff
@@ -1,4 +1,10 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
+
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
```
sdelamo, line 6: I think that this should not be necessary. we were applying this in the root build to every submodule. Thus, I have keep it to avoid doing a breaking change accidentally.

security-oauth2/build.gradle
```diff
@@ -1,4 +1,10 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
+
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
```
sdelamo, line 6: I think that this should not be necessary. we were applying this in the root build to every submodule. Thus, I have keep it to avoid doing a breaking change accidentally.

security-session/build.gradle
```diff
@@ -1,10 +1,13 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
```
sdelamo, line 5: I think that this should not be necessary. we were applying this in the root build to every submodule. Thus, I have keep it to avoid doing a breaking change accidentally.

security/build.gradle
```diff
@@ -1,4 +1,10 @@
+plugins {
+    id "io.micronaut.build.internal.module"
+}
+
 dependencies {
+    annotationProcessor "io.micronaut:micronaut-graal"
```
sdelamo, line 6: I think that this should not be necessary. we were applying this in the root build to every submodule. Thus, I have keep it to avoid doing a breaking change accidentally.

melix (approved):
Changes look good to me. One thing you should try to double check everything is fine is to run:

```
# rm -rf build/repo
# ./gradlew pAPTBR
```

and check the result of the repository in `build/repo`.

PR #881, opened 2022-01-02
javadoc: SecurityFilter should mention SecurityFilter instead of JWT

jreznot:
I believe now Micronaut supports many authentication methods and not only JWT.
So SecurityRule must mention SecurityFilter instead.


sdelamo (approved):
Thanks for the contribution

PR #889, opened 2022-01-04
Use new JsonMapper abstraction for serde compatibility

yawkat:
FYI: this is still blocked by a bug in the annotation processor, @graemerocher will fix that and we can merge this once a new snapshot is out


security-oauth2/build.gradle
```diff
@@ -2,6 +2,11 @@ plugins {
     id "io.micronaut.build.internal.module"
 }
 
+repositories {
+    mavenCentral()
```
sdelamo, line 6: `mavenCentral` is not necessary. I think that It is added as a repository by `io.micronaut.build.internal.module`

security/build.gradle
```diff
@@ -2,6 +2,11 @@ plugins {
     id "io.micronaut.build.internal.module"
 }
 
+repositories {
+    mavenCentral()
```
sdelamo, line 6: `mavenCentral` is not necessary. I think that It is added as a repository by `io.micronaut.build.internal.module`

sdelamo (changes requested):
This fails with: 

```
% ./gradlew -x dependencyUpdates check --parallel --continue

> Task :security:compileJava
warning: unknown enum constant When.MAYBE
  reason: class file for javax.annotation.meta.When not found
/Users/sdelamo/github/micronaut-projects/micronaut-security/security/src/main/java/io/micronaut/security/authentication/ServerAuthentication.java:37: error: Type to deserialize as [io.micronaut.security.authentication.ClientAuthentication], must be a subtype of the annotated type: io.micronaut.security.authentication.ServerAuthentication
public class ServerAuthentication implements Authentication {
```

security-oauth2/build.gradle
```diff
@@ -13,6 +18,8 @@ dependencies {
     implementation "io.micronaut:micronaut-http-client"
     compileOnly "io.micronaut:micronaut-session"
 
+    annotationProcessor "io.micronaut.serde:micronaut-serde-processor:1.0.0-SNAPSHOT"
```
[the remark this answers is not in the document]
sdelamo: I've done this change.

security/build.gradle
```diff
@@ -14,6 +19,8 @@ dependencies {
 
     compileOnly "io.micronaut:micronaut-management"
 
+    annotationProcessor "io.micronaut.serde:micronaut-serde-processor:1.0.0-SNAPSHOT"
```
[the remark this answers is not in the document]
sdelamo: I've done this change.

security-oauth2/build.gradle
sdelamo: I've done this change.

security/build.gradle
sdelamo: I've done this change.

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/OpenIdClientFactory.java
```diff
@@ -66,15 +70,16 @@
...
 
-    private final ObjectMapper objectMapper;
+    private final JsonMapper jsonMapper;
 
     /**
      * @param beanContext The bean context
-     * @param objectMapper Object Mapper
+     * @param jsonMapper Object Mapper
      */
-    OpenIdClientFactory(BeanContext beanContext, ObjectMapper objectMapper) {
```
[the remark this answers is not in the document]
yawkat: the constructor is package-private, should be fine like this

yawkat:
> This fails with:
> 
> ```
> % ./gradlew -x dependencyUpdates check --parallel --continue
> 
> > Task :security:compileJava
> warning: unknown enum constant When.MAYBE
>   reason: class file for javax.annotation.meta.When not found
> /Users/sdelamo/github/micronaut-projects/micronaut-security/security/src/main/java/io/micronaut/security/authentication/ServerAuthentication.java:37: error: Type to deserialize as [io.micronaut.security.authentication.ClientAuthentication], must be a subtype of the annotated type: io.micronaut.security.authentication.ServerAuthentication
> public class ServerAuthentication implements Authentication {
> ```

@graemerocher looks like a bug?

sdelamo:
@yawkat any feedback about this?

graemerocher:
@sdelamo the error should be fixed in the latest snapshot

yawkat:
the error is different now: https://ge.micronaut.io/s/5gde4hmsyj374/failure#1 – missing dependency in serde?

yawkat:
ok the build is fixed, but there is one more issue: In `KeysController` ( https://github.com/micronaut-projects/micronaut-security/blob/3f593f79b78673e203a083129795d7467b0f655f/security-jwt/src/main/java/io/micronaut/security/token/jwt/endpoints/KeysController.java#L67 ), when there are no keys, a map `["keys":[]]` is serialized. For JWKS, this *must* be serialized as `{"keys":[]}`. However with our default jackson settings, this can only work if we call `setSerializationInclusion(JsonInclude.Include.ALWAYS)` on the `ObjectMapper`. This setting can't be accessed with the implementation-independent `JsonMapper` API.

@graemerocher any idea? The best I can think up is transforming the `Map` to a `JsonNode`, which is not subject to serialization inclusion settings, but that'd be quite convoluted.

graemerocher:
perhaps this method needs implementing for serde:

```
    @Override
    default JsonMapper cloneWithFeatures(JsonFeatures features) {
        return this;
    }
```

yawkat:
JsonFeatures has no common API atm, it's basically a placeholder object to store the information of `@JacksonFeatures`

graemerocher:
Right, a common API would need to be added to check if inclusion is enabled

yawkat:
ive added a workaround

PR #894, opened 2022-01-05
fix: basic auth must allow password with :

tchiotludo:
relate to https://github.com/tchiotludo/akhq/issues/939

see https://dev.to/ethanarrowood/what-characters-should-be-allowed-in-http-basic-authentication-userid-and-password-48f0

> Based on this RFC2617 Specification, HTTP Basic Authentication userid can contain any TEXT excluding the symbol :. The password can contain any TEXT. Are these the only rules for Basic Authentication usernames and passwords?


sdelamo:
Thanks for the contribution, I have simplified a little bit the logic. https://github.com/micronaut-projects/micronaut-security/pull/894/commits/65126fa03b0a2379a944aa0f7d32a31c5685f3f4

PR #899, opened 2022-01-13
Audience client_id validation

mancze:
Current version of the validator returns `true` when:
- audience claim contains multiple values, and 
- `azp` claim is present

regardless whether `client_id` is present in audience or not.

This PR suggest a test covering this case and also fix for it.


sdelamo:
Thanks for your contribution!

PR #941, opened 2022-03-02
Make sensitive endpoints error by default

timyates:
We could go further, and allow users to add some configuration to check custom roles,
but as users have to create a replacement for `SensitiveEndpointRule` themselves, this
is something that can more simply be done in that custom replacement.

It also would make the API change much more breaking.

As this is a change in functionality, I have tagged this as `3.4.0`


graemerocher (approved):

sdelamo (approved):
I updated the endoints section copy which needed to be updated

PR #943, opened 2022-03-03
Fix Sonar issues

alvarosanchez:


security/src/main/java/io/micronaut/security/authentication/DefaultAuthorizationExceptionHandler.java
```diff
@@ -86,16 +85,15 @@ public DefaultAuthorizationExceptionHandler(RedirectConfiguration redirectConfig
...
         }
     }
 
     /**
-     * @param request The request
      * @param exception The exception
      * @return The response to be used when a redirect is not appropriate
      */
-    protected MutableHttpResponse<?> httpResponseWithStatus(HttpRequest request, AuthorizationException exception) {
+    protected MutableHttpResponse<?> httpResponseWithStatus(AuthorizationException exception) {
```
sdelamo, line 96: This is breaking change. Moreover, the request is being passed intentionally to allow people override `httpResponseWithStatus` to have access to it.

security/src/main/java/io/micronaut/security/token/propagation/HttpHeaderTokenPropagatorConfigurationProperties.java
```diff
@@ -42,7 +42,7 @@
     private static final String DEFAULT_PREFIX = HttpHeaderValues.AUTHORIZATION_PREFIX_BEARER;
     private static final String DEFAULT_HEADER_NAME = HttpHeaders.AUTHORIZATION;
 
-    private String prefix = DEFAULT_PREFIX;
+    private String headerPrefix = DEFAULT_PREFIX;
```
sdelamo, line 45: why do we need this? Is not better to have it named as the getters and setters?

security/src/main/java/io/micronaut/security/utils/LoggingUtils.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.micronaut.security.utils;
+
+import org.slf4j.Logger;
+
+/**
+ * Utility methods to avoid verbosity of logging statements. Mostly used to help reduce the Cognitive Complexity of
+ * some methods.
+ */
+public final class LoggingUtils {
```
sdelamo, line 24: should this class be part of core?

security-jwt/src/main/java/io/micronaut/security/token/jwt/bearer/BearerTokenConfigurationProperties.java
```diff
@@ -38,7 +38,7 @@
 
     private boolean enabled = DEFAULT_ENABLED;
     private String headerName = HttpHeaders.AUTHORIZATION;
-    private String prefix = HttpHeaderValues.AUTHORIZATION_PREFIX_BEARER;
+    private String headerPrefix = HttpHeaderValues.AUTHORIZATION_PREFIX_BEARER;
```
sdelamo, line 41: why rename this? Should not be better to keep it with the same name as the getters and setters?

sdelamo (changes requested):

sdelamo:
@alvarosanchez thanks for doing this huge PR. :-]

security/src/main/java/io/micronaut/security/authentication/DefaultAuthorizationExceptionHandler.java
alvarosanchez: I mistakenly removed it. Good catch!

security/src/main/java/io/micronaut/security/token/propagation/HttpHeaderTokenPropagatorConfigurationProperties.java
alvarosanchez: It "cognitively" conflicts with the constant `PREFIX`. Since the constant is public and can't be renamed, I renamed the field.

sdelamo: I see, do you think is worth renaming the constant in the next mayor version and rename the field back to `prefix`?

security-jwt/src/main/java/io/micronaut/security/token/jwt/bearer/BearerTokenConfigurationProperties.java
sdelamo: duplicate of https://github.com/micronaut-projects/micronaut-security/pull/943#discussion_r819199988

security/src/main/java/io/micronaut/security/utils/LoggingUtils.java
alvarosanchez: Potentially. @graemerocher thoughts?

Still, I would like to move forward with it belonging here. We can move it if/when it's part of core.

I'm also annotating it with `@Internal` just in case.

sdelamo: Annotate with `@Internal` in case we move it to core soon.

security-jwt/src/main/java/io/micronaut/security/token/jwt/bearer/BearerTokenConfigurationProperties.java
alvarosanchez: It "cognitively" conflicts with the constant PREFIX. Since the constant is public and can't be renamed, I renamed the field.

security/src/main/java/io/micronaut/security/utils/LoggingUtils.java
alvarosanchez: I am not sure if you added your comment before mine, but I just said that I'll do it. Actually it's committed and pushed now.

sdelamo: We were thinking the same. Thanks!

sdelamo (approved):

security/src/main/java/io/micronaut/security/token/propagation/HttpHeaderTokenPropagatorConfigurationProperties.java
alvarosanchez: I honestly don't think it's worth the effort.

PR #1014, opened 2022-05-17
Correct configuration setting name

dogcow:
It appears that `micronaut.security.redirection.enabled` should be `micronaut.security.redirect.enabled`.


sdelamo:
@dogcow  thanks for the contribution

PR #1022, opened 2022-05-23
fix: IpPatternsRule UNKNOWN if HttpRequest:getRemoteAddress throws

sdelamo:


security/src/main/java/io/micronaut/security/rules/IpPatternsRule.java
```diff
@@ -79,27 +79,32 @@ public Publisher<SecurityRuleResult> check(HttpRequest<?> request, @Nullable Rou
             debug(LOG, "No IP patterns provided. Skipping host address check.");
             return Mono.just(SecurityRuleResult.UNKNOWN);
         } else {
-            InetSocketAddress socketAddress = request.getRemoteAddress();
```
sdelamo, line 82: diff looks ugly but it is just wrapping in a try catch block

timyates (approved):

PR #1049, opened 2022-06-30
WIP: changes to support OpenID Connect RP-Initiated Logout 1.0 

wetted:
see: https://openid.net/specs/openid-connect-rpinitiated-1_0.html

closes Issue #1045 

This is a work in progress and not necessarily a correct solution.

PR #1053, opened 2022-07-12
Get keycloak working, but test still fails

timyates:
I made some changes to the keycloak setup for https://github.com/micronaut-projects/micronaut-security/pull/1049

It gets further now, but still fails not sure why 🤔


timyates:
The error is that this doesn't seem to log the user out...

No idea why :-(

https://github.com/micronaut-projects/micronaut-security/blob/80c1fb9391d2d5986b3ce7433a6ebc2e62df2525/security-oauth2/src/test/groovy/io/micronaut/security/oauth2/e2e/OpenIdAuthorizationCodeSpec.groovy#L83

wetted:
> The error is that this doesn't seem to log the user out...
> 
> No idea why :-(

This might be related to a cookie problem @sdelamo noticed when we went over some Keycloak guide work. I think he planned to look into it.

PR #1057, opened 2022-07-22
update to Micronaut 3.5.0

sdelamo:


sdelamo:
@timyates there seems to be a regression `CorsPreflightRequestSpec` fails for Micronaut 3.5.3 and worked in 3.4.4. Can you investigate?

security/src/test/groovy/io/micronaut/security/corspreflightrequest/CorsPreflightRequestSpec.groovy
```diff
@@ -32,4 +39,16 @@ class CorsPreflightRequestSpec extends EmbeddedServerSpecification {
...
     }
+
+    @Requires(property = 'spec.name', value = 'CorsPreflightRequestSpec')
+    @Controller("/register")
+    static class RegisterController {
+
+        @Secured(SecurityRule.IS_AUTHENTICATED)
+        @Delete
+        @Produces(MediaType.TEXT_PLAIN)
+        def register() {
```
sdelamo, line 50: use `String` not `def`

timyates: Still don't see why, but ok

PR #1063, opened 2022-07-26
Make OpenIdAuthenticationMapper.createAuthenticationResponse reactive…

### LINKED ISSUE #927: Make OpenIdAuthenticationMapper.createAuthenticationResponse reactive
### Feature description

 Currently, the OpenIdAuthenticationMapper interface method "createAuthenticationResponse" does not return a Publisher.

Upon receiving an ID token from an OpenID provider, the application may want to retreive user permissions or attributes from other resources, like a database. The application would need to block in order to retreive those before adding them to the AuthenticationResponse. It would be helpful for these calls to be reactive and not block the request thread.

maccuaa:
Make the [OpenIdAuthenticationMapper.createAuthenticationResponse](https://github.com/micronaut-projects/micronaut-security/blob/master/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/OpenIdAuthenticationMapper.java#L45) interface method reactive. This brings the interface in line with the [OauthAuthenticationMapper. createAuthenticationResponse](https://github.com/micronaut-projects/micronaut-security/blob/master/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/OauthAuthenticationMapper.java#L57) interface method.

This will allow developers to make non-blocking API calls in the AuthenticationMapper.

Fixes #927


sdelamo:
@maccuaa we will have to do such changes in the next mayor version of Micronaut Security since this is a breaking change.

maccuaa:
@sdelamo No worries, I figured as much.

maccuaa:
I accidentally created this PR from the master branch in my fork which makes it difficult to sync master to create new PRs. I created a new branch with these commits and opened a new PR #1217.

Issue #603, opened 2021-03-25
Slack OAuth


brunodomenici:
Hi,

I'm trying to configure Slack's OAuth provider in my application, but in this provider, we have a parameter (out of OAuth Spec in my opinion) which is the `user_scope` in addition of `scopes` and we can't configure it in oauth config section. 
Without this config, the redirect url generated to Slack misses the parameter `user_scope`. 

Example:

```yaml
    oauth2:
      enabled: true
      default-provider: slack
      clients:
        slack:
          client-id: 'xxx'
          client-secret: 'xxx'
          scopes:
            - identity.email
          user_scope:
            - identity.email
          authorization:
            url: https://slack.com/oauth/v2/authorize
          token:
            url: https://slack.com/api/oauth.v2.access
            auth-method: client-secret-post
```

Slack OAuth doc: https://api.slack.com/authentication/oauth-v2#asking

Any one have facing the same problem with this kind of OAuth provider?

Thank you

sdelamo:
I think this is what you are looking for can be achieved by replacing `DefaultAuthorizationRedirectHandler`: 

```
package example.micronaut;

import io.micronaut.context.annotation.Replaces;
import io.micronaut.http.MutableHttpResponse;
import io.micronaut.security.oauth2.endpoint.authorization.request.AuthorizationRequest;
import io.micronaut.security.oauth2.endpoint.authorization.request.DefaultAuthorizationRedirectHandler;

import javax.inject.Singleton;
import java.util.Map;

@Replaces(DefaultAuthorizationRedirectHandler.class)
@Singleton
public class SlackAuthorizationRedirectHandler extends DefaultAuthorizationRedirectHandler {

    @Override
    protected Map<String, Object> instantiateParameters(AuthorizationRequest authorizationRequest, MutableHttpResponse response) {
        Map<String, Object> parameters = super.instantiateParameters(authorizationRequest, response);
        parameters.remove("scope");
        parameters.put("user_scope", "identity.email"); // https://api.slack.com/scopes
        return parameters;
    }
}
````

Could you please confirm this work for you?

PR #1065, opened 2022-07-27
Add support for requesting additional claims

### LINKED ISSUE #1058: Support for OIDC  "claims" Request Parameter
### Feature description

The current OIDC implementation only supports requesting scopes however some IDPs support requesting additional claims per section 5.5 of the OIDC spec - https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter

This would be nice to have because not all claims are in the standard scope set.

maccuaa:
Update the configuration to allow specifying additional claims to
request in OIDC requests.

Fixes #1058

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfigurationProperties.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfigurationProperties.java
@@ -58,11 +58,13 @@ public class OauthClientConfigurationProperties implements OauthClientConfigurat
             OpenIdScope.EMAIL.toString(),
             OpenIdScope.PROFILE.toString());
     private List<String> defaultScopes = new ArrayList<>();
+    private List<String> defaultClaims = new ArrayList<>();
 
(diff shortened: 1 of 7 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfigurationProperties.java
```diff
@@ -138,6 +140,21 @@ public void setScopes(List<String> scopes) {
...
     }
 
+    @NonNull
+    @Override
+    public List<String> getClaims() {
+        return claims == null ? defaultClaims : claims;
+    }
+
+    /**
+     * Additional claims to request.
```
sdelamo, line 150: This is shown in the documentation. Tell the user that by default is empty.

sdelamo (changes requested):
Thanks for the contribution. Can you add tests?

maccuaa:
Closing PR since this is more easily achieved by extending the DefaultAuthorizationRedirectHandler per the suggestion in #603

Issue #1066, opened 2022-07-28
Security model Serilization not working using Micronaut Serialization


anandjaisy:
### Expected Behavior

Creating a JWT token should serialize the access token

```
private BearerAccessRefreshToken shouldCreateAValidSignedJwtToken(Map<CharSequence, CharSequence> claims) {
        UsernamePasswordCredentials creds = new UsernamePasswordCredentials("admin@local.com", "RockStar.1");
        HttpRequest request = HttpRequest.POST("/login", creds)
                .headers(claims);
        HttpResponse<BearerAccessRefreshToken> rsp = client.toBlocking().exchange(request, BearerAccessRefreshToken.class);
        bearerAccessRefreshToken = rsp.body();
        return bearerAccessRefreshToken;
    }
```

### Actual Behaviour

Access token does not serialize

![image](https://user-images.githubusercontent.com/84489031/181489897-a62b9567-4492-4112-baa8-60468af1aabf.png)

The access token in null, as from the package` io.micronaut.security.token.jwt.render,` it has a dependency of `com.fasterxml.jackson.annotation`

```
import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.annotation.JsonProperty;
import io.micronaut.core.annotation.Introspected;

@Introspected
public class AccessRefreshToken {

    @JsonProperty("access_token")
    private String accessToken;
}
```

Removing `implementation("io.micronaut.serde:micronaut-serde-jackson:1.2.0")` from the project the serialization/deserializtion is working fine.

### Steps To Reproduce

1. clone the repo and run the test `https://github.com/anandjaisy/Miconaut-security-serialization/blob/main/src/test/java/com/example/TestapplicationTest.java` and observe the JWT creation, the access token is null

### Environment Information

1. JDK 17
2. Mac os pro m1

### Example Application

https://github.com/anandjaisy/Miconaut-security-serialization

### Version

micronautVersion=3.5.3

sdelamo:
@timyates can you investigate?

timyates:
Weird, I can reproduce with the attached project.  The BearerToken is correctly serialized to json, but fails to be deserialized correctly

This can be seen (and by way of a reasonable temporary workaround) if you add a local file

```java
@Serdeable
public class BearerTokenReceiver {

    private final String username;
    private final Collection<String> roles;
    @JsonProperty("access_token")
    private final String accessToken;
    @JsonProperty("token_type")
    private final String tokenType;
[...]

PR #1067, opened 2022-07-29
bug: Deserialization of BearerAccessToken with Serde fails

timyates:
As reported in https://github.com/micronaut-projects/micronaut-security/issues/1066

Serializing the BearerAccessToken to JSON works with Serde, but when deserializing
we need to annotate the constructor to use, as otherwise Serde is picking the no-args
one and failing to capture the fields other than username as they are effectively
read-only

A test has been added to micronaut-security here https://github.com/micronaut-projects/micronaut-serialization/pull/252

This PR also rolls back to 3.6.3 and 3.5.x for the core branch (as it is a bugfix)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ projectVersion=3.6.3-SNAPSHOT
 projectGroup=io.micronaut.security
 
 micronautDocsVersion=2.0.0
-micronautVersion=3.5.3
+micronautVersion=3.4.4
 micronautTestVersion=3.4.0
 
 groovyVersion=3.0.10
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-projectVersion=3.7.0-SNAPSHOT
+projectVersion=3.6.3-SNAPSHOT
 projectGroup=io.micronaut.security
 
 micronautDocsVersion=2.0.0
@@ -14,7 +14,7 @@ projectUrl=https://micronaut.io
 githubSlug=micronaut-projects/micronaut-security
 developers=James Kleeh,Sergio del Amo
 
-githubCoreBranch=3.6.x
+githubCoreBranch=3.5.x
 
 bomProperty=micronautSecurityVersion
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ projectGroup=io.micronaut.security
 
 micronautDocsVersion=2.0.0
 micronautVersion=3.4.4
-micronautTestVersion=3.4.0
+micronautTestVersion=3.1.1
 
 groovyVersion=3.0.10
 spockVersion=2.1-groovy-3.0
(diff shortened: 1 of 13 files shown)


gradle.properties
```diff
@@ -14,7 +14,7 @@ projectUrl=https://micronaut.io
 githubSlug=micronaut-projects/micronaut-security
 developers=James Kleeh,Sergio del Amo
 
-githubCoreBranch=3.6.x
+githubCoreBranch=3.5.x
```
sdelamo, line 17: this is a bug fix. lets keep 3.5.x

gradle.properties
```diff
@@ -1,4 +1,4 @@
-projectVersion=3.7.0-SNAPSHOT
+projectVersion=3.6.3-SNAPSHOT
```
sdelamo, line 1: this is a bug fix. lets keep 3.6.3-SNAPSHOT

sdelamo (changes requested):

sdelamo (changes requested):
Can you add the test here:  https://github.com/micronaut-projects/micronaut-serialization/pull/252

Maybe add serde as testImplementation dependency if we don't have it

gradle.properties
timyates: I set it __*Back*__ to 3.5.x?

gradle.properties
timyates: I set it __*Back*__ to 3.6.3?

gradle.properties
sdelamo: never mind. I am not thinking straight.I changed the PR base to 3.6.x

gradle.properties
sdelamo: never mind. I am not thinking straight.I changed the PR base to 3.6.x

timyates:
Test added 👍

sdelamo (approved):

test-suite-serde/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+dependencies {
+    implementation(platform(mn.micronaut.bom))
+    testCompileOnly(mn.micronaut.inject.groovy)
+    testImplementation(libs.spock.core) {
+        exclude module:'groovy-all'
+    }
+    testImplementation(mn.micronaut.test.spock)
+
+    runtimeOnly(libs.logback.classic)
```
sdelamo, line 16: this should be `testRuntimeOnly`

test-suite-serde/build.gradle
```diff
@@ -0,0 +1,38 @@
+plugins {
+    id 'groovy'
+    id 'java-library'
+}
+repositories {
+    mavenCentral()
+}
+dependencies {
+    implementation(platform(mn.micronaut.bom))
```
sdelamo, line 9: this should be `testImplementation`

sdelamo (changes requested):

PR #1075, opened 2022-08-18
test: ignore if testcontainers system property is set to false

sdelamo:
./gradlew -Dtestcontainers=false test


vignjatov (approved):
lgtm

Issue #1091, opened 2022-09-13
Investigate/address deprecation comment in `OauthClientConfiguration`


wetted:
### Issue description

The `io.micronaut.security.oauth2.configuration.OauthClientConfiguration.getClientCredentials()` method has a "deprecated" inline comment in the code, but is not officially marked as `@Deprecated`. 

```
    default Optional<ClientCredentialsConfiguration> getClientCredentials() {
        return Optional.empty(); //@deprecated Default implementation will be removed in next mayor version.
    }
```

Decide what we need to do about it for Micronaut 4.

PR #1093, opened 2022-09-13
ci: version to 4.0.0 Micronaut 4.0.0-SNAPSHOT githubCoreBranch=4.0.x …

wetted:
…remove deprecations


wetted:
This still has some broken tests to address, as well as #1091

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/propagation/ClientCredentialsHttpClientFilter.java
```diff
@@ -76,7 +76,7 @@ public ClientCredentialsHttpClientFilter(OutgoingHttpRequestProcessor outgoingHt
                                              BeanContext beanContext) {
         this.outgoingHttpRequestProcessor = outgoingHttpRequestProcessor;
         this.oauthClientConfigurationCollection = oauthClientConfigurationStream
-                .filter(config -> config.getClientCredentials().map(ClientCredentialsConfiguration::isEnabled).orElse(false))
```
sdelamo, line 79: why this change?

sdelamo:
@wetted re-request a review once the build passes. Let me know if you need help.

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/propagation/ClientCredentialsHttpClientFilter.java
wetted: That's an artifact of when I started looking into deprecating the method referred to by #1091. I didn't mean to push that change. I have reverted that for now.

sdelamo (approved):

PR #921, opened 2022-01-31
BREAKING CHANGE: Use the ErrorProcessor API for Auth Exceptions

timyates:
Closes #918

PR #1095, opened 2022-09-16
Deprecate constructors to be removed in the next major version

timyates:
See https://github.com/micronaut-projects/micronaut-security/pull/921

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/authorization/response/AuthorizationErrorResponseExceptionHandler.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/authorization/response/AuthorizationErrorResponseExceptionHandler.java
@@ -33,7 +33,7 @@ public class AuthorizationErrorResponseExceptionHandler implements ExceptionHand
     /**
      * Default constructor.
      *
-     * @deprecated This will be removed in the next major version.
+     * @deprecated This will be removed in the next major version, so that this class uses the ErrorProcessor API
      */
     @Deprecated
     public AuthorizationErrorResponseExceptionHandler() {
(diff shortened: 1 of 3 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/authorization/response/AuthorizationErrorResponseExceptionHandler.java
```diff
@@ -30,6 +30,16 @@
 @Singleton
 public class AuthorizationErrorResponseExceptionHandler implements ExceptionHandler<AuthorizationErrorResponseException, MutableHttpResponse<?>> {
 
+    /**
+     * Default constructor.
+     *
+     * @deprecated This will be removed in the next major version.
```
sdelamo, line 36: Tell the users the handler will be refactor to use the ErrorProccessor API

security/src/main/java/io/micronaut/security/authentication/AuthenticationExceptionHandler.java
```diff
@@ -39,6 +39,11 @@
 public class AuthenticationExceptionHandler implements ExceptionHandler<AuthenticationException, MutableHttpResponse<?>> {
     protected final ApplicationEventPublisher<?> eventPublisher;
 
+    /**
+     * @param eventPublisher The event publisher
+     * @deprecated This will be removed in the next major version.
+     */
+    @Deprecated
```
sdelamo, line 46: Tell the users the handler will be refactor to use the ErrorProccessor API

sdelamo (changes requested):

sdelamo (approved):

PR #1107, opened 2022-10-04
Support for Keycloak v17+ in AuthorizationServer::infer

adamkobor:
Keycloak 17+ doesn't have the `/auth/` fragment in its pre-set URLs, therefore the original logic in `AuthorizationServer` is outdated, and it's not possible to infer the right type of newer Keycloak instances based on the URL of the issuer. The biggest consequence of this is that `EndSessionEndpointResolver` is not able to register the right end session endpoint for Keycloak.


adamkobor:
@sdelamo could you let the GitHub Actions run, please? In this case I could see if the build pass, at least, and fix the necessary stuff until the review starts. Thanks 🙏

sdelamo:
I will check this pr soon.

anuragdeshpande:
Any updates on this please, I am running into issues on end-session endpoints too. Keycloak is now in v20.

adamkobor:
Just a side note: this one alone won't solve the issue with Keycloak and RP initiated logouts, https://github.com/micronaut-projects/micronaut-security/pull/1049 should solve the latter one

PR #1113, opened 2022-10-14
Fixed serialization of ServerAuthentication when running in a native image

msupic:
In the authentication process, the `io.micronaut.security.session.SessionLoginHandler` saves an authentication to a session. When Redis is used as a session storage, the authentication is serialized before sent to Redis. The serialization of ServerAuthentication doesn't work as expected when running in a native image (the serialized result doesn't contain any value from ServerAuthentication instance). The test project can be found here: https://github.com/msupic/server-auth/blob/main/src/main/java/server/auth/HomeController.java. Adding
```
@TypeHint(value = {ServerAuthentication.class}, accessType = {TypeHint.AccessType.ALL_PUBLIC})
```
solves the serialization issue when running in a native image

### DIFF
--- a/security/src/main/java/io/micronaut/security/authentication/ServerAuthentication.java
+++ b/security/src/main/java/io/micronaut/security/authentication/ServerAuthentication.java
@@ -18,6 +18,7 @@ package io.micronaut.security.authentication;
 import com.fasterxml.jackson.annotation.JsonValue;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.annotation.TypeHint;
 import io.micronaut.security.token.config.TokenConfiguration;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -33,6 +34,7 @@ import java.util.Map;
  * @author James Kleeh
  * @since 3.0.0
  */
+@TypeHint(value = {ServerAuthentication.class}, accessType = {TypeHint.AccessType.ALL_PUBLIC})
 public class ServerAuthentication implements Authentication {
 
     private static final String JSON_KEY_NAME = "name";


alvarosanchez:
@sdelamo if I'm not mistaken, `@TypeHint` is meant to be used for third-party classes. In this case, we could simply add `@ReflectiveAccess` to `ServerAuthentication`. Or perhaps even `@Introspected` that should fix the issue too.

sdelamo:
@timyates can you create a [`test-suite-graal` module like the one you did for core](https://github.com/micronaut-projects/micronaut-core/pull/8280).

The following test from @msupic works for JIT. I would like to see it fail for naitve. 

```java
package server.auth;

import io.micronaut.http.MediaType;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;
import io.micronaut.http.annotation.Produces;
import io.micronaut.json.JsonObjectSerializer;
import io.micronaut.security.annotation.Secured;
import io.micronaut.security.authentication.ServerAuthentication;
import io.micronaut.security.rules.SecurityRule;

import java.util.HashMap;
import java.util.Map;
import java.util.Optional;

@Controller
public class HomeController {

    private JsonObjectSerializer jsonObjectSerializer;

    public HomeController(JsonObjectSerializer jsonObjectSerializer) {
        this.jsonObjectSerializer = jsonObjectSerializer;
    }

    @Secured(SecurityRule.IS_ANONYMOUS)
    @Produces(MediaType.TEXT_PLAIN)
    @Get("/serialize")
    String serialize() {
        Map<String, Object> attributes = new HashMap<>();
        attributes.put("testKey", "testValue");

        ServerAuthentication auth = new ServerAuthentication("testName", null, attributes);

        Optional<byte[]> resultOpt = jsonObjectSerializer.serialize(auth);

        return new String(resultOpt.get());
    }
}

```
```java
package server.auth;

import io.micronaut.http.HttpRequest;
import io.micronaut.http.client.BlockingHttpClient;
import io.micronaut.http.client.HttpClient;
import io.micronaut.http.client.annotation.Client;
import io.micronaut.runtime.EmbeddedApplication;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.Assertions;

import jakarta.inject.Inject;

import static org.junit.jupiter.api.Assertions.assertEquals;

@MicronautTest
class ServerAuthTest {

    @Inject
    @Client("/")
    HttpClient httpClient;


    @Test
    void foo() {
        BlockingHttpClient client = httpClient.toBlocking();
        assertEquals("{\"name\":\"testName\",\"attributes\":{\"testKey\":\"testValue\"}}", client.retrieve(HttpRequest.GET("/serialize"), String.class));
    }

}
```

sdelamo:
@msupic this is fixed via https://github.com/micronaut-projects/micronaut-security/pull/1139

PR #1123, opened 2022-10-27
fix: prepend context path for redirections

sdelamo:
Currently, the redirection URLs don't work if the user defines server context path. 

This creates a new API `RedirectService` to provide redirection url which include server context path if present.


security/src/main/java/io/micronaut/security/config/SecurityConfigurationProperties.java
```diff
@@ -40,6 +40,12 @@ public class SecurityConfigurationProperties implements SecurityConfiguration {
      */
     @SuppressWarnings("WeakerAccess")
     public static final boolean DEFAULT_ENABLED = true;
+
+    /**
+     * The default enable value.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public static final boolean DEFAULT_INTERCEPT_URL_MAP_PREPEND_PATTERN_WITH_CONTEXT_PATH = false;
```
sdelamo, line 48: I have default this to false to avoid a breaking change

security/src/main/java/io/micronaut/security/config/SecurityConfiguration.java
```diff
@@ -53,4 +53,12 @@ default AuthenticationStrategy getAuthenticationProviderStrategy() {
     default boolean isRejectNotFound() {
         return true;
     }
+
+    /**
+     *
+     * @return Whether the intercept URL patterns should be prepended with context path if defined.
```
timyates, line 59: Add `@Since`

security/src/main/java/io/micronaut/security/config/InterceptUrlMapConverter.java
```diff
@@ -40,12 +43,31 @@ public class InterceptUrlMapConverter implements TypeConverter<Map<String, Objec
...
+
+    /**
+     * @param conversionService     The conversion service
+     * @param serverContextPathProvider context path provider
+     * @param interceptUrlMapPrependPatternWithContextPath Whether the intercept URL patterns should be prepended with context path if defined.
+     */
+    @Inject
+    InterceptUrlMapConverter(ConversionService conversionService,
+                             ServerContextPathProvider serverContextPathProvider,
+                             @Value("${" + SecurityConfigurationProperties.PREFIX + ".intercept-url-map-prepend-pattern-with-context-path:" + SecurityConfigurationProperties.DEFAULT_INTERCEPT_URL_MAP_PREPEND_PATTERN_WITH_CONTEXT_PATH + "}") boolean interceptUrlMapPrependPatternWithContextPath) {
```
timyates, line 67: Why does this not just inject the SecurityConfiguration?

timyates (approved):

timyates:
3 questions, but I don't think they need to block anything

timyates:
We also have test failures, and I'm not sure why

security/src/main/java/io/micronaut/security/config/InterceptUrlMapConverter.java
sdelamo: There is an instantiation error. I think it is because there is a circular dependency. `SecurityConfigurationProperties` the implementation of `SecurityConfiguration` uses the `InterceptUrlMapConverter`

sdelamo:
> We also have test failures, and I'm not sure why

There seems to be an bug with Servlet. I have worked around it by injecting a collection. But I am going to send a PR to servlet.

PR #1131, opened 2022-11-09
fix: decouple converter from config

### LINKED ISSUE #1130: Micronaut CORS configuration does not expose the specified header(s)
### Expected Behavior

When exposing the `Location` header via the CORS configuration in `application.yml`, I expect the header to be accessible in front-end scripts.

### Actual Behaviour

When exposing the `Location` header via the CORS configuration in `application.yml`, the header is NOT accessible in front-end scripts.

### Steps To Reproduce

1. Clone the repository
2. Start up a Keycloak authentication server with the given `realm-export.json` file
3. Start up the Micronaut application
4. Open the given `index.html` file in your browser
5. See that the `Location` header is NOT logged to the console
6. If you change the version to 3.7.2 and update `micronaut-security-jwt` and `micronaut-security-oauth2` to use 3.8.0, I can see the `Location` header

### Environment Information

- Operating System: macOS Ventura 13.0
- JDK (I assume this issue would still occur with JDK 17): openjdk 20-ea 2023-03-21
OpenJDK Runtime Environment (build 20-ea+21-1545)
OpenJDK 64-Bit Server VM (build 20-ea+21-1545, mixed mode, sharing)

### Example Application

https://github.com/lbkulinski/micronaut-3.7.3-regression

### Version

3.7.3

sdelamo:
Fixes #1130

### DIFF
--- a/security-oauth2/src/test/groovy/io/micronaut/security/oauth2/configuration/CorsConfigurationSpec.groovy
+++ b/security-oauth2/src/test/groovy/io/micronaut/security/oauth2/configuration/CorsConfigurationSpec.groovy
@@ -1,0 +1,26 @@
+package io.micronaut.security.oauth2.configuration
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.http.server.HttpServerConfiguration
+import io.micronaut.json.convert.JsonConverterRegistrar
+import spock.lang.Issue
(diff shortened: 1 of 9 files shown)


yawkat:
i'm going to add the test case i built to find this bug

yawkat (approved):

Issue #604, opened 2021-03-25
Okta OpenID Connect Configuration causing target registration timeouts in AWS ECS


elliott-pope-smarsh:
### Steps to Reproduce

1. Create an HTTP API protected with Okta OpenID Connect Configuration and the `/health` endpoint enabled
`build.gradle`
```
implementation("io.micronaut.security:micronaut-security-oauth2")
implementation("io.micronaut.security:micronaut-security-jwt")
```
`application.yml`
```
micronaut:
  security:
    authentication: idtoken
    oauth2:
      clients:
        okta:
          client-id: <client id>
          client-secret: <client secret>
          openid:
            issuer: <issuer URL>
endpoints:
  health:
    enabled: true
    sensitive: false
    details-visible: AUTHENTICATED
```
2. Deploy that on a `t2.micro` instance on AWS ECS
3. Configure an Application Load Balancer to call the `/health` endpoint to register the target

### Expected Behaviour

Container should become healthy and target should be registered

### Actual Behaviour

Call times out and the target fails to register. The logs show 
```
[default-nioEventLoopGroup-1-2] ERROR io.micronaut.http.server.netty.RoutingInBoundHandler - Unexpected error occurred: Error instantiating bean of type [io.micronaut.security.oauth2.client.DefaultOpenIdProviderMetadata]: Read Timeout
```

After some discussion on the Gitter, I think that this call to Okta is blocking the `default` event loop which prevents the server from responding to requests. I was able to resolve the issue by configuring separate client and server event loops of size 10
```
micronaut:
  server:
    netty:
      worker:
        event-loop-group: server-event-loop
  netty:
    event-loops:
      server-event-loop:
        num-threads: 10
      client-event-loop:
        num-threads: 10
  http:
    client:
      event-loop-group: client-event-loop
```
but this shouldn't be necessary

### Environment Information

- **Operating System**: AWS ECS Optimized Linux
- **Micronaut Version:** 2.4.1
- **JDK Version:** 15

### Example Application

I'll work on a sample application so that this can be more easily tested and verified. The application that this issue is coming from is proprietary

sdelamo:
Please, update a sample app if you can. 

Without doing changes to the event loop could you try also to increase Micronaut HTTP Client read timeout. 

`http.client.read-timeout`

graemerocher:
increasing the read timeout is not the solution to this. Need to review why the okta config is blocking the event loop

graemerocher:
[...]

PR #1140, opened 2022-11-16
Fetch OpenID Connect metadata with @Contex and @parallel  don't block the netty event loop

sdelamo:
OpenID connect metadata fetching should not be done in netty event loop thread.

See: https://github.com/micronaut-projects/micronaut-security/issues/604


timyates (approved):

graemerocher:
`@Parallel` is experiemental, in this case it is probably ok since concurrent access to this bean doesn't happen?

graemerocher (approved):

PR #1151, opened 2022-11-25
Add kid to EC signature header

### LINKED ISSUE #1150: Add kid to the jwt token header for EC signing
### Feature description

It looks like we are not adding the kid to the header for EC signed tokens.

https://github.com/micronaut-projects/micronaut-security/blob/master/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/ec/ECSignatureGenerator.java#L63

Unless there is a reason behind that choice this would be quite useful.

For example, we are adding the kid for RSA signed jwt
https://github.com/micronaut-projects/micronaut-security/blob/master/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/rsa/RSASignatureGenerator.java#L76

yoeluk:
add kid to header for ec signatures

Closes  #1150


security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/ec/ECSignatureGenerator.java
```diff
@@ -44,6 +48,13 @@ public class ECSignatureGenerator extends ECSignature implements SignatureGenera
     public ECSignatureGenerator(ECSignatureGeneratorConfiguration config) {
```
sdelamo, line 48: I think it would be better to add a method `default Optional<String> getKid() { return Optional.empty() }` to `ECSignatureGeneratorConfiguration` instead.

sdelamo (changes requested):

yoeluk:
ended up forking the project again... the new PR is now pointing to 3.9.x

PR #1159, opened 2022-11-28
add kid to jwt header for EC signature

yoeluk:
add kid to the jwt header

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/ec/ECSignatureGenerator.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/ec/ECSignatureGenerator.java
@@ -22,6 +22,7 @@ import com.nimbusds.jose.crypto.ECDSASigner;
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.security.token.jwt.signature.SignatureGeneratorConfiguration;
 import java.security.interfaces.ECPrivateKey;
 
@@ -37,6 +38,7 @@ public class ECSignatureGenerator extends ECSignature implements SignatureGenera
 
     private final ECPrivateKey privateKey;
 
+    @Nullable
     private final String kidId;
 
     /**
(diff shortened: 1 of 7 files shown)


yoeluk:
Closes #1150

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/ec/ECSignatureGenerator.java
```diff
@@ -37,6 +38,14 @@ public class ECSignatureGenerator extends ECSignature implements SignatureGenera
 
     private ECPrivateKey privateKey;
 
+    /**
+     *
+     * @return the key id
+     */
+    public Optional<String> getKid() {
```
sdelamo, line 45: annotate the return type as `@NonNull`. Add `@since` to the Javadoc

sdelamo (changes requested):
Thanks for the PR. Could you add a test?

sdelamo (approved):

yoeluk:
@sdelamo yeah sure, I am okay with that. while I was looking at it I draw the conclusion that the current behaviour, when we don't set the kid, of the underlying library (Jose) will set the kid id as `null` so that for consistence and compile-time performance, the `null` was best than optional but otherwise it is fine with me to us an optional too.
For reference this is the build() implementation on the JWSHeader.Builder

```
public JWSHeader build() {

	return new JWSHeader(
		alg, typ, cty, crit,
		jku, jwk, x5u, x5t, x5t256, x5c, kid, b64,
		customParams, parsedBase64URL);
}
```
so if the kid isn't set it will be null anyways.

sdelamo:
> @sdelamo yeah sure, I am okay with that. while I was looking at it I drew the conclusion that the current behaviour, when we don't set the kid, of the underlying library (Jose) will set the kid id as `null` so that for consistency and compile-time performance, the `null` was best than optional, but otherwise it is fine with me to us an optional too. For reference, this is the build() implementation on the JWSHeader.Builder
> 
> ```
> public JWSHeader build() {
> 
> 	return new JWSHeader(
> 		alg, typ, cty, crit,
> 		jku, jwk, x5u, x5t, x5t256, x5c, kid, b64,
> 		customParams, parsedBase64URL);
> }
> ```
> 
> so if the kid isn't set it will be null anyways.

Yup, we are still passing null or a `kid` to the Nimbus Jose Header builder.  However, for our API I think an `@NonNull Optional<String>` is better than a `@Nullable String getKid()`. But a small detail of course.

PR #1171, opened 2022-12-08
Avoid initializing logging at build time for GraalVM

graemerocher:
Custom conditions are encoding into the AnnotationMetadata which is initialized at build time, for this reason conditions can never use logging safely otherwise they will initialize the log infrastructure. This PR moves  the logging in `AuthenticationModeCondition` to a static inner class which avoids initializing the logging when the class is initialized.


sdelamo:
Should not we backport this to 3.9.x to make it work for MN 3?

PR #1184, opened 2022-12-21
fix: Cleaned up and fixed logic in JwtValidator

jan-leila:
made JwtTokenValidator signature and encryption object names match JwtTokenGenerator

fixed only the first encryption config being used to try and decrypt tokens removed redundant verification code
fixed key id sorting logic
simplified code


sdelamo:
@jan-leila the proposed change is a breaking change.

PR #1193, opened 2023-01-05
Enable aot

sdelamo:


melix (approved):

timyates:
@melix Any idea what's causing the failure?  I assumed it was the public aot plugin pulling in old dependencies, but I couldn't fix it with substitutions 😞

melix:
Which failure are you talking about? Because there are many. If that's the failure about not being able to resolve some dependencies, it may be because of the change from the core BOM to the platform BOM.

timyates:
So just running `./gradlew :s-aot:check`, I get multiple possible `InterceptorRegistry` bean candidates

https://ge.micronaut.io/s/6iaddsbndjsqc/console-log?page=1#L180

From my experiance, this is usually due to multiple versions of micronaut-aop being on the classpath

And indeed, `security-aot` has aop 2.5 on in the annotationProcessor dependencies https://ge.micronaut.io/s/6iaddsbndjsqc/dependencies?dependencies=aop&expandAll&focusedDependency=WzMsMCw1ODcsWzIsMCxbMTNdXV0

Could this cause it?  Do annotationProcessor dependencies end up on the test classpath?

Not sure what's pulling in 2.5 either 🤔

timyates:
It seems to be

```
    id("io.micronaut.build.internal.aot-module")
```

pulling in the old version 🤔

melix:
> Could this cause it? Do annotationProcessor dependencies end up on the test classpath?

The annotation process dependencies do not end up on the runtime classpath (test or not). You can see it in the build scan. It is possible that having aop 2.5 on the `annotationProcessor` path is the cause, though. It comes transitively via `inject`: https://ge.micronaut.io/s/6iaddsbndjsqc/dependencies?focusedDependency=WzMsMCw1ODcsWzMsMCxbNTY2LDU4N11dXQ&toggled=W1szXSxbMywwXSxbMywwLFs1NjZdXV0

If we look at the build script, we see that 2 plugins are applied: https://github.com/micronaut-projects/micronaut-security/blob/enable-aot/security-aot/build.gradle.kts#L2-L3

The 1st one is the security internal module, which doesn't seem to do anything suspicious.

The 2d one is the `micronaut-build` AOT plugin, which sets up the project in order to be able to build AOT modules. This plugin does _not_ add anything to the `annotationProcessor` path, but it applies the base module transitively.

This plugin will read the `micronautVersion` property, and also the version from the catalog. In both cases, it should be `4.0.0-SNAPSHOT`.

After looking closely, we can see that the 2.5.0 version comes transitively from `micronaut-docs`: https://ge.micronaut.io/s/4ybfol6q54q2q/dependencies?toggled=W1sxXSxbMSwwXSxbMSwwLFsxN11dLFsxLDAsWzE2XV1d

So a workaround would be to declare the BOM in annotation processor:

```
dependencies {
    annotationProcessor(platform(libs.micronaut.core.bom))
}
```

which then yields: https://ge.micronaut.io/s/fv4ulr7muc42g/dependencies?toggled=W1sxXSxbMSwwXV0

timyates:
@graemerocher @dstepanov When we re-enable the AOT module (and add the test back in), we get an error

```
Distributed Test Run :security-aot:test > Partition 1 in session 1 on localhost-executor-1 > OpenIdProviderMetadataFetcherCodeGeneratorSpec verify OpenIdProviderMetadataFetcherCodeGenerator generates OpenIdProviderMetadataFetcher per openid client FAILED

  io.micronaut.context.exceptions.BeanInstantiationException: Error instantiating bean of type  [io.micronaut.serde.support.DefaultSerdeIntrospections]

  Message: Multiple possible bean candidates found: [InterceptorRegistry, InterceptorRegistry]
```

https://ge.micronaut.io/s/osvthaiic6lti/console-log?task=:security-aot:test

I seem to recall this is caused by having two different versions of micronaut-aop on the test runtime classpath, however I cannot see that

https://ge.micronaut.io/s/osvthaiic6lti/dependencies?dependencies=aop&expandAll&toggled=W1syLDNdLFsyLDJdLFsyLDFdLFsyLDBdLFsxXSxbMF0sWzNdLFs0XSxbNV0sWzZdLFs3XSxbOF0sWzldLFsxMF0sWzExXSxbMTJdLFsxM10sWzE0XSxbMTVdLFsxNl0sWzE3XV0

Do you have any idea what's wrong?

As an aside, a way of showing where beans were created in the case there are multiple in error would be awesome 😉

timyates:
Raised https://github.com/micronaut-projects/micronaut-core/issues/8581

dstepanov:
@timyates @sdelamo Can you resolve the conflicts?

timyates:
Updated the versions as well...  I suspect the error will go away...  CI will tell...

timyates:
@sdelamo running `./test-suite-aot.sh`, the second optimized test is failing as 

```
Message: No bean of type [io.micronaut.security.oauth2.client.OpenIdProviderMetadataFetcher] exists.
```

Not sure what has gone wrong at the moment...  Any ideas?

PR #1202, opened 2023-01-17
fix: multiple possible candidates for DefaultOpenIdProviderMetadata

sdelamo:
Close: #1199

### DIFF
--- a/security-oauth2/src/test/groovy/io/micronaut/security/oauth2/client/DefaultOpenIdClientSpec.groovy
+++ b/security-oauth2/src/test/groovy/io/micronaut/security/oauth2/client/DefaultOpenIdClientSpec.groovy
@@ -1,0 +1,81 @@
+package io.micronaut.security.oauth2.client
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.core.util.CollectionUtils
+import io.micronaut.core.util.StringUtils
+import io.micronaut.http.HttpRequest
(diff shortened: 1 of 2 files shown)


security-oauth2/src/test/groovy/io/micronaut/security/oauth2/client/DefaultOpenIdClientSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+import io.micronaut.security.rules.SecurityRule
+import spock.lang.Issue
+import spock.lang.Specification
+
+class DefaultOpenIdClientSpec extends Specification {
+
+    @Issue("https://github.com/micronaut-projects/micronaut-security/issues/1199")
+    void "DefaultOpenIdClient is created correctly"() {
+        given:
+        EmbeddedServer google = ApplicationContext.run(EmbeddedServer, ["spec.name": "DefaultOpenIdClientSpecGoogle"])
```
sdelamo, line 24: how beautiful is Micronaut Framework that you can create three servers in the same test seamlessly.

timyates (approved):
Ahhhh, because we weren't passing the metadata to createBean as an argument, it was trying to grab a single one out of the context, and failing as there were two?

sdelamo:
> Ahhhh, because we weren't passing the metadata to createBean as an argument, it was trying to grab a single one out of the context, and failing as there were two?

correct. We did not have any test which reproduced such an scenario.

PR #1237, opened 2023-02-28
feat: builder of DefaultOpenIdProviderMetadata

sdelamo:
This PR adds a builder for `DefaultOpenIdProviderMetadata` and uses the builder in the AOT code generation.

In Micronaut Security 4.0, I want to remove the setters from DefaultOpenIdProviderMetadata, and the builder is helpful in avoiding calling a constructor with too many parameters from the AOT code.


timyates (approved):
Thanks for the PR description!  So Phase 2 will be to remove setters, and add a private constructor?

PR #1249, opened 2023-03-13
Move to jakarta packages.

altro3:
Fixes for gradle version catalog.
Removed unnecessary libs from catalog.
micronaut-views 4.0.0-SNAPSHOT


altro3:
@graemerocher Hmm... I tried to figure out what was wrong with this test, but I couldn't. I can say for sure that this is not related to the transition to jakarta or the update of com.sun.activation:javax.activation to org.eclipse.angus:angus-activation. I think I need help :-(

graemerocher:
might be related to secrets

sdelamo:
Thanks for the PR @altro3 

However, I have encapsulated the move from `javax.validation` to `jakarta.validation` in its own PR  #1250  to avoid mixing several things in the same PR.

altro3:
@graemerocher @sdelamo I have only one question about the testcontainers library: IDEA shows me that this library is already in one of the version directories, which is very logical and correct in my opinion, but if I remove these versions from the libs.versions.toml file, then gradle does not find their.

In general, I would like to know: is there a version of testcontainers in the micronaut-test versions directory and if so. why can't you use it.

By the way, I think that selenium should also be present in the micronaut test version directory

altro3:
The same question about `libs.junit.jupiter.engine.` Why I can't use it from other versions catalog?

altro3:
@graemerocher @sdelamo Found problem - that was my mistake here:
изображение

sdelamo:
Thanks of the PR. Incorporated your changes via: #1258 #1257 #1256 #1253 #1252 #1251 #1259

altro3:
@sdelamo thanks! But you lost some changes. See this: https://github.com/micronaut-projects/micronaut-security/pull/1260

Issue #1287, opened 2023-04-02
Wrong cookie path if server context path is set to '/'


skobow:
### Expected Behavior

Cookie path should be issued correctly even though the server context path is configured to be '/'.

### Actual Behaviour

When server context path is configured to be '/' the path of the issued cookie is set to '//' and therefore authentication fails as the cookie is not send with requests to '/'.
With Micronaut 3.8.6 it works even with context configured.

### Steps To Reproduce

add the following config

```
server:
    context-path: ${CONTEXT_PATH:`/`}
```

and check the cookies path in your browsers developer tools. It should be set to '//'

### Version

3.8.8

skobow:
After further investigation I found out that this issue is caused be the changes of commit #1277. I am unsure whether context path configuration of '/' should be considered to be invalid configuration or not. Either way it should be handled accordingly.

As the default context path equals to '/' I would tend to not consider it to be invalid when explicitly configured.

sdelamo:
Micronaut Security 3.9.5 addresses this issue.

PR #1288, opened 2023-04-03
fix: handle explicitly configured default context path (#1287)

skobow:


sdelamo (approved):
Thanks I have added an extra test https://github.com/micronaut-projects/micronaut-security/pull/1288/commits/5cf1aa577c8848404c4719665819af05e7fc8517

Issue #1222, opened 2023-02-09
AuthenticationResponse throws 200 status code even if username and password are wrong


rgogulakrishnan:
### Expected Behavior

Should throw 401 status. 

### Actual Behaviour

Sends 200 status code. Looks like AuthenticationExceptionHandler is not being invoked for failure or error cases. 

### Steps To Reproduce

1. Create an app with security-jwt and reactor features. 
2. Implement AuthUsernamePasswordProvider as mentioned in guide. 
3. Run app
4. Make a login request with wrong username and password. 

```
curl --request POST \
  --url http://localhost:8080/login \
  --header 'Content-Type: application/json' \
  --data '{
	"username": "sherlock",
	"password": "invalid"
}'
```

### Environment Information

Ubuntu and Windows
JDK17

### Example Application

https://github.com/rgogulakrishnan/mn-auth-issue

### Version

3.8.4

rgogulakrishnan:
Updated example project.

rgogulakrishnan:
Any update on this? Is this a bug?

tuantrannav:
according to the javadoc:

A 303 HTTP Response or 200 HTTP Response if loginFailure is null, for example if RedirectConfiguration is disabled.

It seems if you dont enable redirect it will respond 200 OK. I wonder if it should return 401 instead.

sdelamo:
When you set `micronaut.security.authentication` to  `cookie` you enable `JwtCookieLoginHandler` which extends from `CookieLoginHandler`. 

The failure case when you don't want a redirect current returns a 200 intentionally. We have tests for that. @rgogulakrishnan a nd @tuantrannav  I agree with you that maybe a 401 maybe more idiomatic for the non redirect scenario. This may consider a breaking change. Thus, I am going to send the change to master which is the branch which targets Micronaut Framework 4. 

In the meanwhile if you want you can change the behaviour by providing a replacement

Add such a class to your project:

```java
package com.example;

import io.micronaut.context.annotation.Replaces;
import io.micronaut.core.annotation.Nullable;
import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.MutableHttpResponse;
import io.micronaut.security.authentication.AuthenticationResponse;
import io.micronaut.security.config.RedirectConfiguration;
import io.micronaut.security.config.RedirectService;
import io.micronaut.security.errors.PriorToLoginPersistence;
[...]

PR #1323, opened 2023-05-19
BREAKS: change Status for no redirect and  cookie

sdelamo:
For applications with `micronaut.security.authentication` set to `cookie` and `micronaut.security.redirect.enabled` set to `false`, the server responds with 401 HTTP Status code instead of 200 for login failed attempts.

see https://github.com/micronaut-projects/micronaut-security/issues/1222

### DIFF
--- a/security/src/main/java/io/micronaut/security/token/cookie/CookieLoginHandler.java
+++ b/security/src/main/java/io/micronaut/security/token/cookie/CookieLoginHandler.java
@@ -107,7 +107,7 @@ public abstract class CookieLoginHandler implements RedirectingLoginHandler<Http
     public MutableHttpResponse<?> loginFailed(AuthenticationResponse authenticationFailed, HttpRequest<?> request) {
         try {
             if (loginFailure == null) {
-                return HttpResponse.ok();
+                return HttpResponse.unauthorized();
             }
             URI location = new URI(loginFailure);
             return HttpResponse.seeOther(location);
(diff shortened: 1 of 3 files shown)


sdelamo:
ping @jeremyg484

security/src/main/java/io/micronaut/security/token/cookie/CookieLoginHandler.java
```diff
@@ -107,7 +107,7 @@ public MutableHttpResponse<?> loginSuccess(Authentication authentication, HttpRe
     public MutableHttpResponse<?> loginFailed(AuthenticationResponse authenticationFailed, HttpRequest<?> request) {
```
jeremyg484, line 107: The JavaDoc for this method needs updating to reflect the new behavior.

jeremyg484 (changes requested):
Looks good other than the one needed JavaDoc update, and seems like the more appropriate behavior.

Are you thinking we should have OpenRewrite preserve the old behavior? I can see how we could do it by generating a replacement `LoginHandler` to do so...but is that old behavior worth preserving?

sdelamo:
> Looks good other than the one needed JavaDoc update, and seems like the more appropriate behavior.
> 
> Are you thinking we should have OpenRewrite preserve the old behavior? I can see how we could do it by generating a replacement `LoginHandler` to do so...but is that old behavior worth preserving?

No, I don't think this should be target by OpenRewrite.

security/src/main/java/io/micronaut/security/token/cookie/CookieLoginHandler.java
sdelamo: good catch. I have updated the javadoc

jeremyg484 (approved):

PR #1328, opened 2023-05-25
Security + Tomcat fails with fails No active propagation context

sdelamo:


sdelamo:
@dstepanov is this something you could help with?

graemerocher:
have you tried the latest snapshot / build of servlet?

sdelamo:
> have you tried the latest snapshot / build of servlet?

latest upgrade to servlet works

PR #1365, opened 2023-06-30
Enable binary compatibility check

wetted:


buildSrc/src/main/groovy/io.micronaut.build.internal.security-module.gradle
```diff
@@ -2,3 +2,9 @@ plugins {
     id "io.micronaut.build.internal.security-base"
     id "io.micronaut.build.internal.module"
 }
+
+micronautBuild {
```
sdelamo, line 6: You don't need to do this, it is enabled by default

sdelamo (changes requested):

Issue #1374, opened 2023-07-10
NPE in SecurityFilter.logAuthenticationAttributes


kolisko:
### Expected Behavior

prevent NPE when `entry.getValue()` is null

### Steps To Reproduce

when `authentication.getAttributes()` entry has `null` value for any key

### Environment Information

all

### Version

all

PR #1375, opened 2023-07-10
fix: prevent NPE in SecurityFilter

kolisko:
close: #1374


sdelamo:
thanks for the contribution @kolisko

Issue #1376, opened 2023-07-10
AbstractClientCredentialsClient cannot recover from errors that occur during client credentials grant type flow


timsearle:
### Expected Behavior

- Given I request a resource that requires client credentials to be attached to the request
- When that client credentials request fails - e.g. ReadTimeoutException or other error
- Then the next time I request the resource, the client credentials flow requests tokens again

Errors would not be cached, only valid tokens from the client credentials flow.

### Actual Behaviour

In the `AbstractClientCredentialsClient`, any errors that occur during the client credentials flow are "cached" in the Flux. This means subsequent requests to obtain the token, automatically replay the previous errors. Resulting in instant failure:

https://github.com/micronaut-projects/micronaut-security/blob/b0b5299bec0186ff4742fe20dd16d63c36a7c32d/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java#L99

### Steps To Reproduce

See the attached sample repository: https://github.com/timsearle/micronaut-security-issue - we have included non-sensitive sample client credentials, that we may remove in the future, feel free to replace with your own ID/secret/token URL.

**Note: To reproduce the issue, we have configured a client read-timeout of 10ms.**

Execute the `OAuthBugTest`, examine the test logs and identify that the initial request attempted an OAuth flow, whereas the subsequent attempt, then immediately finishes with the cached error from the original request.

```
    15:08:25.735 [Test worker] INFO  i.m.context.env.DefaultEnvironment - Established active environments: [test]
15:08:25.935 [Test worker] DEBUG i.m.s.o.routes.OauthRouteBuilder - No Oauth controllers found. Skipping registration of routes
15:08:26.063 [Test worker] TRACE i.m.h.server.netty.NettyHttpServer - Binding ouathbug server to *:-1
15:08:26.286 [Test worker] TRACE i.m.s.o.e.t.r.DefaultTokenEndpointClient - Sending request to token endpoint [https://mplanchant.eu.authz.cloudentity.io/mplanchant/default/oauth2/token]
15:08:26.286 [Test worker] TRACE i.m.s.o.e.t.r.DefaultTokenEndpointClient - The token endpoint supports [[client_secret_post]] authentication methods
15:08:26.286 [Test worker] TRACE i.m.s.o.e.t.r.DefaultTokenEndpointClient - Using client_secret_post authentication. The client_id and client_secret will be present in the body
[...]

Issue #991, opened 2022-04-25
ClientCredentials filter cache should not cache errors


ggerbaud:
### Expected Behavior

When using ClientCredentialsClient errors should not be cached, or at least with a TTL.

### Actual Behaviour

Errors are cached and never cleared

For a given scope, the token request Flux is put in the map ([L83](https://github.com/micronaut-projects/micronaut-security/blob/master/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java#L83)). If it's an error, all following calls will ends at [L91](https://github.com/micronaut-projects/micronaut-security/blob/master/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java#L91) even if the error was temporary (e.g. 503).

An easy fix would be to insert this line just before L91
`scopeToPublisherMap.remove(scope);`

### Example Application

https://github.com/ggerbaud/client-credential-issue
The unit test illustrates the issue

### Version

3.x

ggerbaud:
Just add an example application

timsearle:
@ggerbaud @alvarosanchez I've raised a pull request to resolve this issue, please take a look! #1379

PR #1379, opened 2023-07-11
Bug Fix: No longer cache and replay errors in AbstractClientCredentialsClient

timsearle:
Closes https://github.com/micronaut-projects/micronaut-security/issues/1376
Closes https://github.com/micronaut-projects/micronaut-security/issues/991

Please let me know if you need me to target another branch, or if you have any concerns with the approach/tests. We are very motivated to get this merged as soon as we can. Currently, we're patching this class in manually on our service.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java
@@ -81,17 +81,14 @@ public abstract class AbstractClientCredentialsClient implements ClientCredentia
     public Publisher<TokenResponse> requestToken(@Nullable String scope, boolean force) {
         String resolvedScope = scope != null ? scope : NOSCOPE;
         return Flux.from(scopeToPublisherMap.computeIfAbsent(resolvedScope, k -> cachedTokenResponseForScope(scope)))
-                .materialize()
-                .next()
-                .flatMap((Function<Signal<TokenResponse>, Mono<TokenResponse>>) tokenNotif -> {
-                    if (!force && tokenNotif.isOnNext() && !isExpired(tokenNotif.get())) {
-                        TokenResponse tokenResponse = tokenNotif.get();
-                        return tokenResponse != null ? Mono.just(tokenResponse) : Mono.empty();
-                    } else if (tokenNotif.isOnError()) {
-                        return tokenNotif.getThrowable() != null ? Mono.error(tokenNotif.getThrowable()) : Mono.error(Throwable::new);
-                    }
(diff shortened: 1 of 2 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java
```diff
@@ -81,17 +81,14 @@ public Publisher<TokenResponse> requestToken(@Nullable String scope) {
...
-                    } else if (tokenNotif.isOnError()) {
-                        return tokenNotif.getThrowable() != null ? Mono.error(tokenNotif.getThrowable()) : Mono.error(Throwable::new);
-                    }
-                    return Mono.from(scopeToPublisherMap.computeIfPresent(resolvedScope, (s, tokenResponse) -> cachedTokenResponseForScope(scope)));
-                });
+            .flatMap((Function<TokenResponse, Mono<TokenResponse>>) response -> {
+                if (!force && !isExpired(response)) {
+                    return Mono.just(response);
+                }
+                return Mono.from(scopeToPublisherMap.computeIfPresent(resolvedScope, (s, tokenResponse) -> cachedTokenResponseForScope(scope)));
```
thomasmillercf, line 88: Think you might be able to just do a map here.

timsearle: `cachedTokenResponseForScope` is asynchronous, so flatMap is appropriate.

thomasmillercf:
Looks good, we should get this in

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java
CraftyFella: I'm with @thomasmillercf both returns are wrapping the result in a Mono.from.. so wouldn't this work with a map and return the Mono.from?

sdelamo (approved):
Thanks a lot for the contribution

security-oauth2/src/main/java/io/micronaut/security/oauth2/client/clientcredentials/AbstractClientCredentialsClient.java
timsearle: Either would work, but the code in the closure can result in a new asynchronous request for a token if it is expired or forced. It is generally best practice to use flatMap for non-blocking scenarios, and it will ensure the new Mono is auto subscribed. If this was a synchronous block, e.g. just transforming the data, I'd use a map.

CraftyFella: It's the Mono.from that forces the flatMap.. for some reason I saw Mono.just and read the 2nd one as Mono.just.. what you have is correct.

PR #1402, opened 2023-08-01
Feature/add external cache

Riciardos:
https://github.com/micronaut-projects/micronaut-security/discussions/1403

The idea is mainly for serverless contexts where you want short startup times and not having to retrieve the JWKS from the OpenID Connect provider every time.
The 'ExternalJwksCache' interface can be implemented in different ways, I've done it in 2 ways, with DynamoDb and Redis. For Redis you need `micronaut-redis` on your classpath, for DynamoDb I've just added the AWS SDK for now but I think that can be changed to `micronaut-aws` probably? There is a null implementation by default, which just does nothing.

Both implementations just store the JWKS as string with the provider url as key. There is a default TTL to both of 1 hour.
For DynamoDb you can specify the tablename, what the tables PK/SK keys are named etc.


Riciardos:
Not fully fledged yet with test code coverage etc. Discussion for it is here: https://github.com/micronaut-projects/micronaut-security/discussions/1403

sdelamo:
@Riciardos have you tried our Micronaut Security Ahead of time optimizations? https://micronaut-projects.github.io/micronaut-security/latest/guide/#aot They help exactly with this.

Riciardos:
Hi @sdelamo , ah I did not see that actually, that's great. Though these keys rarely change, is there any concern around having to recompile when they do? I see there is a warning about that in that section, so might still be worth adding something like this in for an 'inbetween' solution?

I'm happy for this to be closed if there isn't a need for it.

sdelamo:
@Riciardos I am going to close this for now. We may revaluate in the future. Thanks for the PR.

PR #1459, opened 2023-10-05
Alter Javadoc in TokenGenerator.java

baylorpaul:
Fix incorrect time unit in Javadoc. The comment on "expiration" was indicating the value is in "milliseconds", but this value is in "seconds" in io.micronaut.security.token.jwt.generator.JwtTokenGenerator


sdelamo (approved):
Thanks!

PR #1500, opened 2023-11-17
add default implementation method TokenResolver::resolveTokens

sdelamo:
I should have added it here: https://github.com/micronaut-projects/micronaut-security/pull/1498


timyates (approved):
The default is calling a deprecated method?

Ahhh, for things compiled against an old version?

Issue #1048, opened 2022-06-30
Exception loading JWK URL (behind a proxy corporate)


elfelli:
### Expected Behavior

Validate a JWT behind a proxy corporate 

### Actual Behaviour

I'm trying to configure AKHQ (0.21.0) based on micronaut v3.0.1 with okta oauth2. (okta is accessible only with a proxy corporate )

But i got a TimeoutException when trying to load JWKS URL behind the proxy 

micronaut:
  http:
    client:
      proxy-type: HTTP
      proxy-address: proxy.company.net:8080

https://github.com/micronaut-projects/micronaut-security/blob/963eabd8bcae21f81ed3679cb0c3215abdf65846/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java#L166

it seem that the **load function** is not using the proxy config micronaut.http.client....
Authenticate against authentication providers is OK with proxy config 
JWT generation is OK ( token is valid i checked it manually)
But micronaut security fail when trying to validate JWT

**a work around** is to override jwks-uri: https://company.okta-emea.com/oauth2/aus2wqdfqfqkr0i7/v1/keys to mirror the json response under a direct access url without proxy 

TRACE LOG

9m Starting health monitor check
[...]

PR #1509, opened 2023-11-29
Use Micronaut HTTP client for remote JWK Set validation

jeremyg484:
This needs some further general cleanup (haven't completed all of the JavaDoc updates, for example) and I want to add some additional tests, but I wanted to go ahead and get some early feedback before I do that additional work.

The implementation works for replacing the internal `JWKSet` URL fetching that uses JDK Http APIs with use of the Micronaut HTTP client, so that the client settings can be fully configured (to solve problems such as #1048).

My initial hesitation is around the changes that had to be made in order to propagate a JWT service name that can be matched up with the HTTP client service name, in the same manner as is done in `DefaultTokenEndpointClient`. I'm open to suggestion as to ways to improve this. Most of the impact is to internal classes and protected methods, so perhaps this is all ok.      

I did retain the existing `JwkSetFetcher#fetch` method as a potential fallback, though at the moment it will not be used and would probably need to be marked as deprecated.


security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwkSetFetcher.java
```diff
@@ -38,6 +39,15 @@ public interface JwkSetFetcher<T> {
     @Blocking
     Optional<T> fetch(@Nullable String url);
```
sdelamo, line 40: should this be deprecated?

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignature.java
```diff
@@ -147,12 +148,13 @@ public boolean verify(SignedJWT jwt) throws JOSEException {
 
     /**
      * Instantiates a JWKSet for a given url.
+     * @param providerName The name of the JWKS configuration.
      * @param url JSON Web Key Set Url.
      * @return a JWKSet or null if there was an error.
      */
     @Nullable
-    protected JWKSet loadJwkSet(String url) {
```
sdelamo, line 154: this is a breaking change

security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/JwksSignatureConfiguration.java
```diff
@@ -27,6 +27,9 @@
  */
 public interface JwksSignatureConfiguration {
 
+    @NonNull
+    String getName();
```
sdelamo, line 31: Extend from the `Named` interface instead.

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/validation/DefaultOpenIdTokenResponseValidator.java
```diff
@@ -147,33 +147,36 @@ protected Optional<JWT> validateClaims(@NonNull OauthClientConfiguration clientC
...
                 .validate(openIdTokenResponse.getIdToken(), null);
     }
 
     /**
-     *
+     * @param clientConfiguration The OAuth 2.0 client configuration
      * @param openIdProviderMetadata The OpenID provider metadata
      * @return A {@link JwksSignature} for the OpenID provider JWKS uri.
      */
-    protected JwksSignature jwksSignatureForOpenIdProviderMetadata(@NonNull OpenIdProviderMetadata openIdProviderMetadata) {
```
sdelamo, line 170: this is a breaking change.

sdelamo (changes requested):
It is going in the right direction. 

However,  We cannot have breaking changes. Moreover, I think it would to have a `JwksClient` API where we can encapsulate the whole `HttpClient` instantiation etc.  

We could do a `@EachBean(JwksSignatureConfiguration.class`) and create a `JwksClient` bean. 

It needs test.

jeremyg484:
With the latest changes I have:

- Completed refactoring to retain (and deprecate) the existing API methods in favor of the new ones that support propagation of the provider name. The existing methods will still function by passing `null` for the provider name in which case they use the default HTTP client.

- Made the http client a compile only dependency and provided two implementations of a new JwksClient interface - one that uses the Nimbus library's `ResourceRetriever` API that was being used here before, and one that uses the Micronaut `HttpClient`. The `HttpClient` implementation takes precedent if the `HttpClient` is on the class path. This might be overkill, but I think it's the best way to ensure that existing apps don't break.

- Added a set of tests that simulate usage with a proxy for both implementations of JwksClient and ensures that the configured `HttpClient` settings get used. The non-proxy default settings use case is already thoroughly covered with the existing tests.

Some remaining TODOs:

- Figured out why the Java 21 CI build is failing
- Check the AOT optimizations and modify as necessary
- Add some examples to the docs about configuring the `HttpClient` settings for JWK Set retrieval.

jeremyg484:
@sdelamo Ready for review now. Note that in this final set of changes that I've moved away from deprecating methods in `DefaultOpenIdTokenResponseValidator` (see my prior comments) and instead added a non-serialized name property to `DefaultOpenIdProviderMetadata`. This approach simplified the changes that had to be made to the AOT optimizations.

PR #1523, opened 2023-12-06
Jwks micronaut http client improvements

sdelamo:
I have changed the `JwksClient` API from: 

```
String load(@Nullable String providerName, @NonNull String url);`
```

to: 

```java
@SingleResult
Publisher<String> load(@Nullable String providerName, @NonNull String url);
```

to block at the very last moment. 

I have annotated constructors etc with `@Deprecated(forRemoval = true, since = "4.5.0")`

I made the none http-client implementation of `JwksClient` secondary and allowed the possibility to disable `HttpClientJwksClient` via `micronaut.security.token.jwk.signatures.jwks-client.http-client.enabled=false`


security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ResourceRetrieverJwksClient.java
```diff
@@ -31,22 +35,24 @@
...
             Resource resource = resourceRetriever.retrieveResource(new URL(url));
-            return resource.getContent();
+            return Mono.just(resource.getContent());
         } catch (IOException e) {
             if (LOG.isErrorEnabled()) {
                 LOG.error("Exception loading JWK from " + url, e);
             }
+            return Mono.empty();
         }
-        return null;
```
jeremyg484, line 50: `resourceRetriever.retrieveResource` is a blocking call, so I think it should be wrapped in `Mono.fromCallable`:
```suggestion
        return Mono.fromCallable(() -> {
            try {
                return resourceRetriever.retrieveResource(new URL(url));
            } catch (IOException e) {
                if (LOG.isErrorEnabled()) {
                    LOG.error("Exception loading JWK from " + url, e);
                }
                return null;
            }
        }).map(Resource::getContent).subscribeOn(Schedulers.boundedElastic());
```

sdelamo: resolved by https://github.com/micronaut-projects/micronaut-security/pull/1523/commits/3dd723c14d830adfa946b70c7c8f741a5c673724

PR #1524, opened 2023-12-06
doc: Custom binding of authenticated user

sdelamo:
Close: #1430

### DIFF
--- a/src/main/docs/guide/retrievingAuthenticatedUser/customAuthenticatedUser.adoc
+++ b/src/main/docs/guide/retrievingAuthenticatedUser/customAuthenticatedUser.adoc
@@ -1,0 +1,14 @@
+
+You can create a custom argument binder to bind the authenticated user to a custom class tailored to your application needs.
+
+If, in your application, the authenticated user has an email address, you can create a class such as:
+
+snippet::io.micronaut.security.docs.customauthentication.AuthenticationWithEmail[]
+
+and then a `TypedRequestArgumentBinder`:
+
+snippet::io.micronaut.security.docs.customauthentication.AuthenticationWithEmailArgumentBinder[]
+
+Then you can bind it in a controller method parameter:
+
+snippet::io.micronaut.security.docs.customauthentication.CustomAuthenticationTest[tags="method"]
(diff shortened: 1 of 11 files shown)


jeremyg484 (changes requested):
While I think showing this particular example is useful in its own right, I don't really think it addresses the specific use case from #1430 wherein they are fully replacing the default `Authentication` object such that a call to `request.getUserPrincipal(AuthenticationWithEmail.class)` would be expected to work. 

To address that case, I think we'd want to also show that it's possible to just extend `AbstractPrincipalArgumentBinder` as they did, but explain how that approach only works with something like a custom `AuthenticationProvider`.

Also as an aside that should probably be addressed separately, I notice that providing a custom implementation of `TokenValidator` as you have in the tests here seems like a nice simple way to enhance the `Authentication` attributes, but I don't see it mentioned anywhere in the documentation.

jeremyg484:
> They are not fully replacing authentication. They are creating:
> 
> ```java
> class TokenAuthentication implements Authentication {
> String token
> }
> ```
> 
> I think this PR adds the docs they would need for they usecase (bind such a class in a controller method parameter).

In https://github.com/micronaut-projects/micronaut-security/issues/1430#issuecomment-1700018632 they said that simply extending `AbstractPrincipalArgumentBinder` solved their problem.

i.e., in Java their example would be:

```
@Singleton
public class TokenAuthenticationArgumentBinder extends AbstractPrincipalArgumentBinder<TokenAuthentication> {

    public TokenAuthenticationArgumentBinder() {
        super(TokenAuthentication.class);
    }
}
```

The `bind` method in their implementation is thus going to call `httpRequest.getUserPrincipal(TokenAuthentication.class)`. For that to be working successfully, they have to have somehow replaced the `Authentication` that is stored in the request, whether with a custom `AuthenticationProvider`, a custom `OpenIdAuthenticationMapper`, etc. Thus the example binder shown in this PR is more than what they need for that specific case.

jeremyg484 (approved):
See my prior comment above. I think this is good in its own right to be merged, but I don't think it addresses the specific case in #1430, so I don't think it should close that issue. 

We should address that separately, either with my suggested enhancement in #1522, or by providing a separate documentation example of extending `AbstractPrincipalArgumentBinder`, or both.

PR #1526, opened 2023-12-07
Add Imperative AuthenticationProvider and Reactive AuthenticationProvider

### LINKED ISSUE #1476: add non reactive API for AuthenticationProvider
### Issue description

To simplify the usage of Micronaut Security for users who don't want to use reactive code

sdelamo:
Close #1476


jeremyg484:
@sdelamo It won't let me actually request you as reviewer since you started the PR.

This is ready for review, but a few things to consider right off the top:

As you will see, I realized one caveat of this approach to adapting the `BlockingAuthenticationProvider` implementations via `@EachBean` is that it requires the addition of a qualifier to each implementation if providing more than one. I don't love that, but do you think providing multiple implementations would be common enough to warrant a different approach?

Also, I've only provided the Java implementation of the example in the docs - that's all we have for `AuthenticationProvider` as well. I'm thinking we should go ahead and provide Groovy and Kotlin versions of both as usual - I'll look at that first thing tomorrow.

sdelamo (commented):
> As you will see, I realized one caveat of this approach to adapting the BlockingAuthenticationProvider implementations via @EachBean is that it requires the addition of a qualifier to each implementation if providing more than one. I don't love that, but do you think providing multiple implementations would be common enough to warrant a different approach?

I think it is fine to require a Name qualifier. I have changed the `BlockingAuthenticationProvider` to extend `io.micronaut.core.naming.Named` and the docs show now the example. 

> Also, I've only provided the Java implementation of the example in the docs - that's all we have for AuthenticationProvider as well. I'm thinking we should go ahead and provide Groovy and Kotlin versions of both as usual - I'll look at that first thing tomorrow.

Yes, we should provide kotlin and groovy examples always even if previous snippet of the docs did not. I have [provided Groovy and Kotlin](https://github.com/micronaut-projects/micronaut-security/pull/1532/files) tests.

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param blockingAuthenticationProvider An instance of {@link BlockingAuthenticationProvider} to be adapted
+     * @return An {@link AuthenticationProvider} adapted from the blocking provider
+     * @param <T> The request type
+     */
+    @EachBean(BlockingAuthenticationProvider.class)
+    <T> AuthenticationProvider<T> createAuthenticationProvider(BlockingAuthenticationProvider<T> blockingAuthenticationProvider) {
+        return new BlockingAuthenticationProviderAdapter<>(blockingAuthenticationProvider, scheduler);
+    }
+
+    private static class BlockingAuthenticationProviderAdapter<T> implements AuthenticationProvider<T> {
```
sdelamo, line 62:
```suggestion
    private static final class BlockingAuthenticationProviderAdapter<T> implements AuthenticationProvider<T> {
```

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.concurrent.ExecutorService;
+
+/**
+ * A factory for adapting {@link BlockingAuthenticationProvider} beans to expose them as {@link AuthenticationProvider}.
+ *
+ * @since 4.5.0
+ */
+@Factory
+@Internal
+class BlockingAuthenticationProviderFactory {
```
dstepanov, line 37: Can we not call it blocking? It doesn't need to block

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,79 @@
...
+        private final Scheduler scheduler;
+
+        private BlockingAuthenticationProviderAdapter(BlockingAuthenticationProvider<T> blockingAuthenticationProvider, Scheduler scheduler) {
+            this.blockingAuthenticationProvider = blockingAuthenticationProvider;
+            this.scheduler = scheduler;
+        }
+
+        @Override
+        public Publisher<AuthenticationResponse> authenticate(T httpRequest, AuthenticationRequest<?, ?> authenticationRequest) {
+            return Mono.fromCallable(() -> blockingAuthenticationProvider.authenticate(httpRequest, authenticationRequest))
```
dstepanov, line 75: I would expect to use the new API directly in the filter, not have it wrapped in a reactive

dstepanov: Preferable this change should allow not engage any reactive filtering when all the filters are non-reactive

dstepanov (changes requested):

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
sdelamo: I have changed the API `ImperativeAuthenticationProvider`. 

If the user implementation is blocking he should add the `@Blocking` annotation to the overridden method.

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
sdelamo: I think that should be attempted in the next major version of Micronaut Security.

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
dstepanov: What is the plan? Rename it in v5? Or keep it like this forever?
IMHO I would create a new API non-reactive `AuthenticationProvider` and `ReactiveAuthenticationProvider` in a different package and deprecate the old one.

sdelamo: If we were starting from zero, I would not be against  the naming to be `ReactiveAuthenticationProvider` for the publisher api and `AuthenticationProvider` for the imperative API.

But we already have an API in place, which most users will have implemented in their applications and I am not keen to forcing them to change their code for a better name.

Having the same API name in different packages sounds tricky to me as well. 

I don't think having two APIs: the existing `AuthenticationProvider` which is reactive and a new `ImperativeAuthenticationProvider` which is not reactive is too bad.

sdelamo: @dstepanov I have deprecated

io.micronaut.security.authentication.AuthenticationProvider

and created two APIs: 

- Imperative => `io.micronaut.security.authentication.provider.AuthenticationProvider` 
- Reactive => `io.micronaut.security.authentication.provider.ReactiveAuthenticationProvider`

https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/c795eaada8c7253f0a82eef4334195a575273eb8

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
sdelamo: I have removed the factory and I only create `ReactiveAuthenticationProvider` if users define any or if the user authentication provider is blocking https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/c795eaada8c7253f0a82eef4334195a575273eb8

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/password/OauthPasswordAuthenticationProvider.java
```diff
@@ -32,7 +32,7 @@
 import reactor.core.publisher.Flux;
 
 /**
- * An {@link AuthenticationProvider} that delegates to an OAuth 2.0 provider using the
+ * A {@link AuthenticationProvider} that delegates to an OAuth 2.0 provider using the
```
sdelamo, line 35:
```suggestion
 * An {@link AuthenticationProvider} that delegates to an OAuth 2.0 provider using the
```

dstepanov (approved):

sdelamo:
@jeremyg484 can you review?

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
jeremyg484: @sdelamo I like this - it's clean and more consistent with the terminology we use elsewhere (i.e., Micronaut Data).

security/src/main/java/io/micronaut/security/authentication/Authenticator.java
```diff
@@ -45,55 +62,156 @@ public class Authenticator<T> {
...
+        this.securityConfiguration = securityConfiguration;
+        this.imperativeAuthenticationProviders = authenticationProviders;
+        this.scheduler = Schedulers.fromExecutorService(executorService);
+        this.authenticationProviders = Collections.emptyList();
+    }
+
+    /**
+     * @param authenticationProviders A list of available authentication providers
+     * @param securityConfiguration   The security configuration
+     * @deprecated Use {@link Authenticator(List, SecurityConfiguration)} instead.
```
jeremyg484, line 104:
```suggestion
     * @deprecated Use {@link Authenticator#Authenticator(BeanContext, List, List, ExecutorService, SecurityConfiguration)} instead.
```

security/src/main/java/io/micronaut/security/authentication/provider/ReactiveAuthenticationProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.async.annotation.SingleResult;
+import io.micronaut.core.order.Ordered;
+import io.micronaut.security.authentication.Authentication;
+import io.micronaut.security.authentication.AuthenticationRequest;
+import io.micronaut.security.authentication.AuthenticationResponse;
+import org.reactivestreams.Publisher;
+
+/**
+ * Defines a Reactive authentication provider.
```
jeremyg484, line 27:
```suggestion
 * Defines a reactive authentication provider.
```

jeremyg484 (changes requested):
Looks great, just have some nit picking suggestions for JavaDocs.

security/src/main/java/io/micronaut/security/authentication/provider/ReactiveAuthenticationProvider.java
```diff
@@ -0,0 +1,50 @@
...
+     * all authenticators for each authentication request and it is assumed no work
+     * will be done until the publisher is subscribed to.
+     *
+     * @param httpRequest The http request
+     * @param authenticationRequest The credentials to authenticate
+     * @return A publisher that emits 0 or 1 responses
+     */
+    @NonNull
+    @SingleResult
+    Publisher<AuthenticationResponse> authenticate(@Nullable T httpRequest,
```
graemerocher, line 48: rename the `httpRequest` parameter to `requestContext`, since we don't know if this is an HTTP request or not

graemerocher, line 48: since the request is Nullable add a `default` method that just takes the `authenticationRequest` and passes `null` to the `authenticate` method

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProvider.java
```diff
@@ -0,0 +1,47 @@
...
+     * If your implementation is blocking, annotate the overriden method with {@link Blocking} and it will be safely executed on a
+     * dedicated thread in order to not block the main reactive chain of execution.
+     *
+     * @param httpRequest The http request
+     * @param authRequest The credentials to authenticate
+     * @return An {@link AuthenticationResponse} indicating either success or failure.
+     */
+    @NonNull
+    @Executable
+    AuthenticationResponse authenticate(@Nullable T httpRequest, @NonNull AuthenticationRequest<?, ?> authRequest);
```
graemerocher, line 46: rename `httpRequest` to `requestContext` and add `default` method like the other comment

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProviderAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
+
+/**
+ * Adapts between {@link AuthenticationProvider} to {@link ReactiveAuthenticationProvider}.
+ * @param <T> Request
+ */
+@Internal
+public class AuthenticationProviderAdapter<T> implements ReactiveAuthenticationProvider<T> {
```
graemerocher, line 33: does this need to be `public`? Can it be ` final`?

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProviderUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import io.micronaut.inject.ExecutableMethod;
+import io.micronaut.security.authentication.AuthenticationRequest;
+
+import java.lang.annotation.Annotation;
+
+/**
+ * Utility class to check whether {@link AuthenticationProvider#authenticate(Object, AuthenticationRequest)} is annotated with {@link Blocking}.
+ */
+@Internal
+public final class AuthenticationProviderUtils {
```
graemerocher, line 33: does this need to be `public`?

security/src/main/java/io/micronaut/security/authentication/provider/HttpRequestAuthenticationProvider.java
```diff
@@ -0,0 +1,26 @@
...
+package io.micronaut.security.authentication.provider;
+
+import io.micronaut.http.HttpRequest;
+
+/**
+ * {@link AuthenticationProvider} for {@link HttpRequest}.
+ * @author Sergio del Amo
+ * @since 4.5.0
+ */
+public interface HttpRequestAuthenticationProvider extends AuthenticationProvider<HttpRequest>  {
```
graemerocher, line 25: the `HttpRequest` parameter is missing its generic argument type, you should never use raw types

security/src/main/java/io/micronaut/security/authentication/provider/HttpRequestReactiveAuthenticationProvider.java
```diff
@@ -0,0 +1,26 @@
...
+package io.micronaut.security.authentication.provider;
+
+import io.micronaut.http.HttpRequest;
+
+/**
+ * {@link ReactiveAuthenticationProvider} for {@link HttpRequest}.
+ * @author Sergio del Amo
+ * @since 4.5.0
+ */
+public interface HttpRequestReactiveAuthenticationProvider extends ReactiveAuthenticationProvider<HttpRequest>  {
```
graemerocher, line 25: the HttpRequest parameter is missing its generic argument type, you should never use raw types

security/src/main/java/io/micronaut/security/authentication/Authenticator.java
```diff
@@ -45,55 +62,156 @@ public class Authenticator<T> {
...
+     * @param beanContext Bean Context
+     * @param reactiveAuthenticationProviders A list of available Reactive authentication providers
+     * @param authenticationProviders A list of available imperative authentication providers
+     * @param executorService BLOCKING executor service
      * @param securityConfiguration The security configuration
      */
-    public Authenticator(Collection<AuthenticationProvider<T>> authenticationProviders,
+    @Inject
+    public Authenticator(BeanContext beanContext,
+                         List<ReactiveAuthenticationProvider<T>> reactiveAuthenticationProviders,
```
graemerocher, line 89: since you are injecting lists here annotate the `ReactiveAuthenticationProvider` and `AuthenticationProvider` interfaces with `@Indexed` to improvement performance

graemerocher (changes requested):

security/src/main/java/io/micronaut/security/authentication/provider/ReactiveAuthenticationProvider.java
sdelamo: [renamed `httpRquest` parameter to `requestContext`](https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/8001291896502f083133dc31335b240ccc2ba395)

security/src/main/java/io/micronaut/security/authentication/provider/HttpRequestAuthenticationProvider.java
sdelamo: I added generic arguments https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/8c8318dcfcd4595046c8816b75046b2d6541eeb6

security/src/main/java/io/micronaut/security/authentication/provider/HttpRequestReactiveAuthenticationProvider.java
sdelamo: I added generic arguments https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/8c8318dcfcd4595046c8816b75046b2d6541eeb6

security/src/main/java/io/micronaut/security/authentication/Authenticator.java
sdelamo: [I annotated both with `@Indexed`](https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/c737dd155138cd26f01ed3e9ab2fb3b9a1d4acfd)

graemerocher (approved):

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProviderUtils.java
sdelamo: I moved it to the same package as `Authenticator` to make it package private. 

https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/69c09f7bae7680b0e2b9ec0d64f7c72d79a0790a

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProviderAdapter.java
sdelamo: I moved it to the same package as Authenticator to make it package private.

https://github.com/micronaut-projects/micronaut-security/commit/69c09f7bae7680b0e2b9ec0d64f7c72d79a0790a

security/src/main/java/io/micronaut/security/authentication/provider/ReactiveAuthenticationProvider.java
sdelamo: Added a default method https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/e3a5d1bd8fea5f0a571e9e49f895e664b7ac2683

security/src/main/java/io/micronaut/security/authentication/provider/AuthenticationProvider.java
sdelamo: I added a default method https://github.com/micronaut-projects/micronaut-security/pull/1526/commits/e3a5d1bd8fea5f0a571e9e49f895e664b7ac2683

jeremyg484 (approved):

jeremyg484 (approved):

PR #1528, opened 2023-12-08
Add @CreatedBy and @UpdatedBy annotation

sdelamo:
See [Entity Events](https://micronaut-projects.github.io/micronaut-data/latest/guide/#entityEvents)

This PR adds the `@CreatedBy` and `@UpdatedBy` annotations to populate  some entity fields automatically: 

```groovy
@MappedEntity
class Message {
    @Id
    @GeneratedValue
    @Nullable
    Long id

    @NotBlank
    String title

    @CreatedBy
    @Nullable
    String createdBy
}
```

### DIFF
--- a/security/build.gradle.kts
+++ b/security/build.gradle.kts
@@ -15,6 +15,11 @@ dependencies {
     implementation(mnValidation.micronaut.validation)
     implementation(mnReactor.micronaut.reactor)
 
+    testCompileOnly(mnData.micronaut.data.processor)
+    testImplementation(mnData.micronaut.data.jdbc)
+    testImplementation(mnSql.h2)
+    testImplementation(mnSql.micronaut.jdbc.hikari)
+
     compileOnly(mn.micronaut.http.server)
     compileOnly(mn.micronaut.management)
     compileOnly(mn.jackson.databind)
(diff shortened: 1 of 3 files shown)


sdelamo:
@jeremyg484 I have started the PR, please pick it up from here and ping me when it is ready for review.

sdelamo:
We will need to use the [BeanIntrospection](https://docs.micronaut.io/latest/guide/#introspection) API.

sdelamo:
@jeremyg484 

> One pending thought I have - should we ensure there is a means of mapping from/to full Principal objects of some sort the way that Spring Data does? (See https://docs.spring.io/spring-data/jpa/reference/auditing.html#auditing.annotations)

I think for now it is fine if we only allow a `String` as the type annotated with `@CreatedBy` `@UpdatedBy`. What the user wants to persist is a serialized representation of the user unique id.  Mapping to a fully fledged `Authentication` may mean several queries to a database or a remote service. 

Maybe what we need is an API to allow developers to calculate the user unique id in case they want to use other than `Principal::getName`

```java
@FunctionalInterface
interface UserUniqueIdentifier {
@NonNull String uniqueIdentifier(@NonNull Principal principal);
}
```

and provide a default implementation. 

```java
@Secondary
@Singleton
interface DefaultUserUniqueIdentifier {
@NonNull 
String uniqueIdentifier(@NonNull  Principal principal) {
return principal.getName();
}
}
```

PR #1532, opened 2023-12-12
Blocking authentication provider impr

sdelamo:


security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
```diff
@@ -57,21 +59,14 @@ <T> AuthenticationProvider<T> createAuthenticationProvider(BlockingAuthenticatio
...
-
-        private final BlockingAuthenticationProvider<T> blockingAuthenticationProvider;
-
-        private final Scheduler scheduler;
-
-        private BlockingAuthenticationProviderAdapter(BlockingAuthenticationProvider<T> blockingAuthenticationProvider, Scheduler scheduler) {
-            this.blockingAuthenticationProvider = blockingAuthenticationProvider;
-            this.scheduler = scheduler;
-        }
+    private record BlockingAuthenticationProviderAdapter<T>(
```
jeremyg484, line 62: I considered doing the same (making the class a record) since IntelliJ was suggesting it, but to me this isn't the use case for which records are intended. It's not really just a data carrier, and I don't think it needs the accessor methods that the record provides, etc.

Do you have another reason that I'm not thinking of?

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProvider.java
```diff
@@ -18,13 +18,14 @@
...
 import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.naming.Named;
 
 /**
  * Defines an authentication provider.
  * @since 4.5.0
  * @param <T> Request
  */
-public interface BlockingAuthenticationProvider<T> {
+public interface BlockingAuthenticationProvider<T> extends Named {
```
jeremyg484, line 28: I considered this, but my thought is what value does extending `Named` actually provide? It seems like an unnecessary extra burden on the user in this case. There is no constraint in Micronaut that forces them to use the implemented name in the `@Named` qualifier annotation, is there? Nor does it force them to add the qualifier?

jeremyg484 (changes requested):
Thanks for the assist with the docs! I have a couple minor questions inline.

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProvider.java
sdelamo: > It seems like an unnecessary extra burden on the user in this case. 

I think the burden is minimal

> There is no constraint in Micronaut that forces them to use the implemented name in the @Named qualifier annotation, is there? 
No

> Nor does it force them to add the qualifier?
No

However, together with the docs (where we show using the same name qualifier as in the `getName` method) is friendly reminder for you not to forget to add the name qualifier to the class.

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProviderFactory.java
sdelamo: I have reverted it https://github.com/micronaut-projects/micronaut-security/pull/1532/commits/03ddaf3b34722ba8c4687d71bc2cdc4385058792

security/src/main/java/io/micronaut/security/authentication/BlockingAuthenticationProvider.java
jeremyg484: Ok, fair enough.

jeremyg484 (approved):

PR #1537, opened 2023-12-14
Implementation of `@CreatedBy` and `@UpdatedBy`

jeremyg484:
Initial implementation of support for identity auditing with Micronaut Data through addition of `@CreatedBy` and `@UpdatedBy` annotations.


jeremyg484:
@sdelamo I still need to add docs, but here is my initial implementation if you'd like to go ahead and take a look. I split it out into its own module due to the Micronaut Data dependencies, but we can move it elsewhere if you prefer.

The implementation is based on some helpful support for similar features from the Micronaut Data runtime module (See the usage of `AutoPopulatedEntityEventListener`).

I opted to simply use the `ConversionService` and a default `TypeConverter` implementation for the mapping of `Principal` to `String` - this is easily overridden by the user and would allow for conversion to more complex types if they prefer.

security-audit/src/main/java/io/micronaut/security/audit/event/listeners/UserAuditingEntityEventListener.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
+            getCurrentUserIdentityForProperty(beanProperty).ifPresent(identity -> context.setProperty(beanProperty, identity));
+        }
+    }
+
+    private Optional<Object> getCurrentUserIdentityForProperty(BeanProperty<Object, Object> beanProperty) {
+        return securityService.getAuthentication().flatMap(authentication -> conversionService.convert(authentication, beanProperty.getType()));
+    }
```
graemerocher, line 100: Seems like lots of logic and loops are executed and perhaps do nothing if there is no authentication. Move the check for `securityService.getAuthentication.isPresent()` to an outer most if statement and then loop

graemerocher (changes requested):

security-audit/src/main/java/io/micronaut/security/audit/event/listeners/UserAuditingEntityEventListener.java
jeremyg484: Good point...done.

graemerocher (approved):

sdelamo:
I registered the converter via a `TypeConverterRegistrar` as recommended here: https://github.com/micronaut-projects/micronaut-core/pull/10315

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * An event listener that handles auto-population of entity fields annotated with {@link CreatedBy} or
+ * {@link UpdatedBy} by mapping them from the current {@link Authentication}.
+ *
+ * @author Jeremy Grelle
+ * @since 4.5.0
+ */
+@Requires(classes = { AutoPopulatedEntityEventListener.class, EntityEventContext.class })
+@Singleton
+public class UserAuditingEntityEventListener extends AutoPopulatedEntityEventListener {
```
graemerocher, line 52: can this be `final`? Seems the constructor is default package scope so it it isn't designed to be subclassed. Can it be non-public?

graemerocher (approved):

PR #1539, opened 2023-12-14
Improvements to @CreatedBy

sdelamo:
- Updates listener to get Authentication only once and convert only when necessary. 
- Clear logic about when a property should be persisted in the listener
- Use Java records, groovy POGOs and Kotlin Data classes for documentation 
- Uses Asciidoc ` ann` macro


security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -83,23 +82,30 @@ public boolean preUpdate(@NonNull EntityEventContext<Object> context) {
...
-                if (isUpdate) {
-                    if (!persistentProperty.getAnnotationMetadata().booleanValue(AutoPopulated.class, AutoPopulated.UPDATEABLE).orElse(true)) {
-                        continue;
-                    }
-                }
-
-                final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
-                getCurrentUserIdentityForProperty(beanProperty).ifPresent(identity -> context.setProperty(beanProperty, identity));
-            }
+            securityService.getAuthentication().ifPresent(authentication -> {
```
jeremyg484, line 88: I think at this point the outermost `if (securityService.isAutheticated())` is redundant and can be removed.

jeremyg484 (changes requested):

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
sdelamo: [I have removed the outermost `if`](https://github.com/micronaut-projects/micronaut-security/pull/1539/commits/16f3cd7a9c793e6b88ecf3aae59db2eeac8f170b)

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -83,23 +82,27 @@ public boolean preUpdate(@NonNull EntityEventContext<Object> context) {
...
+                          @NonNull Class<? extends Annotation> listenerAnnotation) {
+        securityService.getAuthentication().ifPresent(authentication -> {
+            Map<Class<?>, Object> valueForType = new HashMap<>();
+            Stream.of(getApplicableProperties(context.getPersistentEntity()))
+                    .filter(persistentProperty -> shouldSetProperty(persistentProperty, listenerAnnotation))
+                    .forEach(persistentProperty -> {
+                        final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
+                        Object value = valueForType.computeIfAbsent(beanProperty.getType(), type -> conversionService.convert(authentication, beanProperty.getType()).orElse(null));
+                        context.setProperty(beanProperty, value);
+                    });
```
graemerocher, line 95: all these streams are worse from a performance perspective

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -83,23 +82,27 @@ public boolean preUpdate(@NonNull EntityEventContext<Object> context) {
...
+    private void populate(@NonNull EntityEventContext<Object> context,
+                          @NonNull Class<? extends Annotation> listenerAnnotation) {
+        securityService.getAuthentication().ifPresent(authentication -> {
+            Map<Class<?>, Object> valueForType = new HashMap<>();
+            Stream.of(getApplicableProperties(context.getPersistentEntity()))
+                    .filter(persistentProperty -> shouldSetProperty(persistentProperty, listenerAnnotation))
+                    .forEach(persistentProperty -> {
+                        final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
+                        Object value = valueForType.computeIfAbsent(beanProperty.getType(), type -> conversionService.convert(authentication, beanProperty.getType()).orElse(null));
+                        context.setProperty(beanProperty, value);
```
jeremyg484, line 94: It occurs to me, this behavior is different from the original - I think we shouldn't be explicitly setting the value to null here if the conversion fails, as it could potentially overwrite an existing stored value. We should probably use `convertRequired` instead.

jeremyg484 (changes requested):

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
sdelamo: I check not to be null before setting the property. I added a test as well https://github.com/micronaut-projects/micronaut-security/pull/1539/commits/a45939f898b47fe2fae46e240324fcfd963fa54a

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
sdelamo: https://github.com/micronaut-projects/micronaut-security/pull/1539/commits/7a05e31642581af3cf1682dca1440b4bf4318e62

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
jeremyg484: @sdelamo Why opt for the null check instead of switching to using `convertRequired`? 

My thinking is that failing to convert if there is a valid `Authentication` to be applied to one of the annotated fields should be an error and brought to the user's attention. It could be that they're using a custom converter and their conversion logic is wrong. Silently ignoring the failure to convert would make such a bug harder to detect.

sdelamo: [I have changed to use `convertRequired`](https://github.com/micronaut-projects/micronaut-security/pull/1539/commits/66aa0ace73a5159da0f3400de328be9883d29a1f)

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -83,23 +87,41 @@ public boolean preUpdate(@NonNull EntityEventContext<Object> context) {
...
 
-                final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
-                getCurrentUserIdentityForProperty(beanProperty).ifPresent(identity -> context.setProperty(beanProperty, identity));
+    @Nullable
+    private Object convert(@NonNull Authentication authentication, @NonNull BeanProperty<Object, Object> beanProperty) {
+        try {
+            return conversionService.convertRequired(authentication, beanProperty.getType());
+        } catch (ConversionErrorException e) {
+            if (LOG.isErrorEnabled()) {
+                LOG.error("cannot convert from {} to {}", authentication.getClass().getSimpleName(), beanProperty.getType(), e);
```
jeremyg484, line 112:
```suggestion
                LOG.error("Cannot convert from {} to {} for bean property {}", authentication.getClass().getSimpleName(), beanProperty.getType(), beanProperty.getName(), e);
```

security/src/main/java/io/micronaut/security/audit/UserAuditingEntityEventListener.java
```diff
@@ -83,23 +87,41 @@ public boolean preUpdate(@NonNull EntityEventContext<Object> context) {
...
+        });
+    }
 
-                final BeanProperty<Object, Object> beanProperty = persistentProperty.getProperty();
-                getCurrentUserIdentityForProperty(beanProperty).ifPresent(identity -> context.setProperty(beanProperty, identity));
+    @Nullable
+    private Object convert(@NonNull Authentication authentication, @NonNull BeanProperty<Object, Object> beanProperty) {
+        try {
+            return conversionService.convertRequired(authentication, beanProperty.getType());
+        } catch (ConversionErrorException e) {
```
jeremyg484, line 110: The next question is, should we actually propagate this error up to `preUpdate` and `prePersist` and then return `false` from them so that the data operation does not proceed?

sdelamo: changed it here: https://github.com/micronaut-projects/micronaut-security/pull/1539/commits/cfa271e10b8e84b8cb49040c6587fe4053fc0e8b

jeremyg484 (approved):
Looks good. I went ahead and cleaned up some unused imports.

PR #1580, opened 2024-02-07
fix: U…Credentials bean-introspection-module=false

sdelamo:
Suppor serialization of UsernamePasswordCredentials for jackson.bean-introspection-module=false


sdelamo:
Regarding the sonar failure: The only new code in this PR is an empty constructor, which is tested with the test in the other module.

timyates (approved):

PR #1582, opened 2024-02-08
fix: AccessRefreshToken bean-introspection-module

sdelamo:


timyates (approved):

timyates:
@sdelamo this sonar bug is right I think https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=1582&id=micronaut-projects_micronaut-security&open=AY2IPRQrZntfTWvBqHU4 do we need to remove the nonnull annotation?

sdelamo:
> @sdelamo this sonar bug is right I think https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=1582&id=micronaut-projects_micronaut-security&open=AY2IPRQrZntfTWvBqHU4 do we need to remove the nonnull annotation?

I could remove the @NonNull annotation. But the API is more correct with the @ NonNull annotations. Those fields should be required. They were also final. 

The problem is that in order to support jackson-databind without bean-introspection-module enabled, I have to add an empty construct.

PR #1589, opened 2024-02-12
fix: don’t expose validation

sdelamo:
Verify it works with [micronaut-hibernate-validator](https://micronaut-projects.github.io/micronaut-hibernate-validator/latest/guide/)

### DIFF
--- a/security-oauth2/build.gradle.kts
+++ b/security-oauth2/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-    id "io.micronaut.build.internal.security-module"
+    id("io.micronaut.build.internal.security-module")
 }
 
 dependencies {
@@ -7,7 +7,8 @@ dependencies {
     annotationProcessor(mnSerde.micronaut.serde.processor)
     annotationProcessor(projects.micronautSecurityAnnotations)
     annotationProcessor(mnValidation.micronaut.validation.processor)
-    api(mnValidation.micronaut.validation)
+    api(mnValidation.validation)
+    testImplementation(mnValidation.micronaut.validation)
     compileOnly(mn.micronaut.inject.java)
     compileOnly(projects.micronautSecurityJwt)
     compileOnly(mn.micronaut.http.server)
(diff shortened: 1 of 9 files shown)


security-oauth2/build.gradle.kts
```diff
@@ -1,13 +1,14 @@
...
+    id("io.micronaut.build.internal.security-module")
 }
 
 dependencies {
     annotationProcessor(mn.micronaut.graal)
     annotationProcessor(mnSerde.micronaut.serde.processor)
     annotationProcessor(projects.micronautSecurityAnnotations)
     annotationProcessor(mnValidation.micronaut.validation.processor)
-    api(mnValidation.micronaut.validation)
+    api(mnValidation.validation) //  // jakarta.validation:jakarta.validation-api
```
timyates, line 10: What's the comment for here?

security/build.gradle.kts
```diff
@@ -10,9 +10,11 @@ dependencies {
     api(mnSerde.micronaut.serde.api) {
         exclude(group = "io.micronaut", module = "micronaut-json-core")
     }
-    api(mnValidation.validation)
+    api(mnValidation.validation) //  // jakarta.validation:jakarta.validation-api
```
timyates, line 13: What's the comment for here?

security-oauth2/build.gradle.kts
sdelamo: it is totally alien to me that `mnValidation.validation` means `jakarta.validation:jakarta.validation-api`.  If instead of`mnValidation.validation` it was `mnValidation.jakarta.validation.api`I would have not added the comment.

security/build.gradle.kts
sdelamo: it is totally alien to me that `mnValidation.validation` means `jakarta.validation:jakarta.validation-api`.  If instead of`mnValidation.validation` it was `mnValidation.jakarta.validation.api`I would have not added the comment.

PR #1613, opened 2024-02-23
SameSite cookie attribute handling on logout

ThomasIommi:
Hi!

I've noticed that in the `TokenCookieClearerLogoutHandler` class, the `logout()` method seems to ignore the configuration of the _SameSite_, _Secure_, and _HTTPOnly_ attributes, focusing only on _Domain_, _Path_, _Name_, and _Max-Age_.

I wonder if this behaviour could deny cross-site authentication proxies from revoking authentication cookies with SameSite=None, Secure, and HTTPOnly attributes, even if originally provided by the proxy itself.

For example, in Chrome v120 I've got exactly replicated this issue, while the logout it's still working in Firefox:
image

I'm not sure if there are specific reasons for the current implementation, but in case the mentioned attributes should also be handled during the logout phase, I propose the attached pull request.

To ensure consistency within the project, I've employed the same method implementation as found in the `TokenCookieLoginHandler` and tried to provide an update that reflects the changes in the unit tests.

I'm new to contributions, I really hope this is an adequate format, I apologize in advance if not, I'm also open for feedbacks and discussions.

Thank you in advance for your time and consideration! :)


sdelamo:
@ThomasIommi thanks for your contribution.

PR #1623, opened 2024-03-04
fix: losing authentication failure reason

sdelamo:


security/src/main/java/io/micronaut/security/authentication/Authenticator.java
```diff
@@ -186,8 +187,7 @@ private AuthenticationResponse authenticate(@NonNull T requestContext,
         }
         return authenticationProviders.stream()
                 .map(provider -> authenticationResponse(provider, requestContext, authenticationRequest))
-                .filter(AuthenticationResponse::isAuthenticated)
-                .findFirst()
+                .min(AUTHENTICATION_RESPONSE_COMPARATOR)
                 .orElseGet(AuthenticationResponse::failure);
```
timyates, line 191: Say we have 3 providers, all responding. with a failure...

Which failure should we be responding with here?

Does `min` mean that the order is unpredictable as it's basically doing a `sort` on the elements?

Instead of this, we could do something like this:

```java
        AtomicReference<AuthenticationResponse> result = new AtomicReference<>(AuthenticationResponse.failure());
        authenticationProviders
            .stream()
            .map(provider -> authenticationResponse(provider, requestContext, authenticationRequest))
            .takeWhile(t -> !result.get().isAuthenticated())
            .forEach(t -> result.updateAndGet(current -> t));
        return result.get();
```

Which I believe keeps the order, but stops when it encounters the first `success`, but will end up returning the last `failure` if there are no successes 🤔

sdelamo: what do you think? https://github.com/micronaut-projects/micronaut-security/pull/1623/commits/2216496b73711f3bdde6b925f6426cce59fbef7f

security/src/main/java/io/micronaut/security/authentication/Authenticator.java
```diff
@@ -184,11 +183,17 @@ private AuthenticationResponse authenticate(@NonNull T requestContext,
         if (securityConfiguration != null && securityConfiguration.getAuthenticationProviderStrategy() == AuthenticationStrategy.ALL) {
             return authenticateAll(requestContext, authenticationRequest, authenticationProviders);
         }
-        return authenticationProviders.stream()
-                .map(provider -> authenticationResponse(provider, requestContext, authenticationRequest))
-                .filter(AuthenticationResponse::isAuthenticated)
-                .findFirst()
-                .orElseGet(AuthenticationResponse::failure);
+        List<AuthenticationResponse> responses = new ArrayList<>();
+        for (AuthenticationProvider provider : authenticationProviders) {
```
timyates, line 187:
```suggestion
        for (AuthenticationProvider<T, ?, ?> provider : authenticationProviders) {
```

For Sonar :-)

timyates (approved):

Issue #1629, opened 2024-03-09
Docs: Token Propagation requires io.micrometer:context-propagation


ArnauAregall:
### Issue description

Reading the reference docs there is no specific mention that Token Propagation feature requires dependency `io.micrometer:context-propagation` dependency [to work properly](https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/token/propagation/TokenPropagationHttpClientFilter.java#L68) (spent some time debugging it).

Also on the samples of the guides, the dependency is not mentioned explicitly but added on all submodules (`gateway`, `userecho`, `intermediate-gateway`, `intermediate-user`).

Opening the issue to reference it in a a PR I'll open in order to add a specific mention for the needed dependency. 🙂

sdelamo:
Thanks for reporting @ArnauAregall we would need add a link to the  [Reactor Context Propagation](https://docs.micronaut.io/latest/guide/#reactorContextPropagation) documentation from the token propagation section.

PR #1630, opened 2024-03-09
doc: Mention io.micrometer:context-propagation on Token Propagation

ArnauAregall:
See  #1629


sdelamo (changes requested):
I think it would be best to link to https://docs.micronaut.io/latest/guide/#reactorContextPropagation

ArnauAregall:
> I think it would be best to link to https://docs.micronaut.io/latest/guide/#reactorContextPropagation

Makes sense, pushed changes (https://github.com/micronaut-projects/micronaut-security/pull/1630/commits/0b7b5c9559ec8313cfe88599786f1b6f1a167169).

sdelamo (approved):

Issue #1634, opened 2024-03-13
`file:` prefix is not working for static jwks configurations


jbreton:
### Expected Behavior

As described in the [documentation](https://micronaut-projects.github.io/micronaut-security/latest/guide/#staticJwks), `file:` prefix should allow to serve a JSON JWKS from anywhere on disk. 

### Actual Behaviour

`file:` prefix tries to load a classpath resource instead of a file on disk.

From the tests I did, if [StaticJwksSignature](security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/StaticJwksSignature.java) uses [ResourceResolver](core/src/main/java/io/micronaut/core/io/ResourceResolver.java) instead of ResourceLoader, `file:` and `classpath:` prefixes work as intended.

### Steps To Reproduce

1. Use a jwks-static configuration using a path starting with `file:`

### Version

4.3.4

PR #1635, opened 2024-03-13
`file:` prefix is not working for static jwks configurations

jbreton:
Fix for #1634 

We found this bug while setting a static Jwks as a workaround of a possible bug regarding a ReadTimeout that could occur when fetching a Jwks from an url.

We are still investigating the cause of the ReadTimeout and haven't opened an issue since it could be caused by our custom JWT validator. Maybe this will ring a bell as it began when we updated to micronaut 4.3.1 and is still occuring in 4.3.4.


sdelamo (changes requested):
Thanks for the contribution. Can you add a test?

jbreton:
@sdelamo It's my first time working with groovy. I looked at other tests to see what I could do. Those tests are far from perfect but it's better than nothing at all.

PR #1637, opened 2024-03-14
Update settings and remove groovy version

timyates:


timyates:
Not sure why this is failing...  I don't think it's this change, I think it's something to do with Keycloak 🤔

timyates:
4.6.x and 4.7.x seem to fail the same way 😞

timyates:
All the jboss/keycloak docker images have been deleted from docker hub

Fix here https://github.com/micronaut-projects/micronaut-security/pull/1640

sdelamo (approved):

Issue #1024, opened 2022-05-25
Support Keycloak 17+


morki:
## Keycloak 17

In Keycloak 17 the default distribution is [powered by Quarkus](https://www.keycloak.org/2022/02/keycloak-1700-released) and default endpoints has changed (removed the `/auth` prefix) so autoconfiguration of `EndSessionEndpoint` is failing.

See https://github.com/micronaut-projects/micronaut-security/pull/1009

## Keycloak 18

In Keycloak 18 there are even [bigger changes](https://www.keycloak.org/2022/04/keycloak-1800-released) as shift to using standard OIDC logout and more. This means the `KeycloakEndSessionEndpoint` is not working anymore.

The behaviour is now following OIDC standard and is working exactly as `OktaEndSessionEndpoint`.

## Proposed solution

Add new configuration property `mode` with following options:
- `auto` (actual behavior using `EndSessionEndpointResolver`, backward compatible, default value)
- `standard` (new behaviour, following the standard, using renamed `OktaEndSessionEndpoint`)

Example:

```yaml
micronaut:
  security:
    oauth2:
      clients:
        internal:
          enabled: true
          client-id: xxx
          client-secret: xxx
          openid:
            issuer: xxx
            end-session:
              enabled: true
              mode: standard # <-- this is the new property
```

## Current workaround for Keycloak 18

We successfuly tricked Micronaut to think that Keycloak is Okta by using `?okta` suffix for issuer URL e.g. `https://sso.xxx.com/realms/master?okta`.

graemerocher:
would you consider sending a PR for this? I think it makes sense to add a separate keycloak18 config so that older versions continue to be supported

morki:
@graemerocher I think the same, what do you think about the proposed solution above? Any naming or other recommendations?

graemerocher:
seems fine, will need to be documented though

sdelamo:
Thanks for the request @morki 

I think we have to support [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html). And I think we have to be able to specify whether you want to use that logout request or a custom.

morki:
New version for this proposal can use property `vendor` instead of `mode`:

```yaml
micronaut:
  security:
    oauth2:
      clients:
        internal:
          enabled: true
          client-id: xxx
          client-secret: xxx
          openid:
            vendor: keycloak-17 # <-- this is the new property
            issuer: xxx
[...]

PR #1640, opened 2024-03-15
Fix Keycloak docker tests for 4.6.x

timyates:
We used to use an elderly version of jboss/keycloak for running our Keycloak tests.

The jboss/keycloak docker image is no longer available, so we need to do something to re-instate the tests.

Instead of using the x86 keycloak/keycloak image, this PR switches to the latest (at time of writing) bitnami/keycloak image. These images have arm versions, so we can get rid of a lot of the complications.

Updating to the latest has raised 2 issues with our current Keycloak support.

1. Keycloak 17+ got rid of the /auth prefix to their URLs.  In this PR I added it back via an environment variable, but we should support some configuration to support this.
2. The logout flow for Keycloak has changed, so we have to enable legacy flow. Basically the redirect_uri parameter has been deprecated, so we need to capture and use some keys to go back to the previous behavior.

~~I will raise issues for these 2 problems.~~ We have an issue for this here https://github.com/micronaut-projects/micronaut-security/issues/1024

This commit also renames a package that was named incorrectly `keycloack`


wetted (approved):

test-suite-geb/src/test/resources/logback.xml
```diff
@@ -7,4 +7,5 @@
     <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
+    <logger name="io.micronaut.security" level="trace"/>
```
sdelamo, line 10:
```suggestion
```

sdelamo (approved):

Issue #1633, opened 2024-03-13
ReadTimeoutException loading JWK causing exponential resource consumption since 4.5.0


etremblay:
### Expected Behavior

When the JWKS endpoint is available, the application load the key information and boot normally.

When the JWKS endpoint is down, the application retry untile the JWKS is up.

### Actual Behaviour

Application fail with 

```
Exception loading JWK
io.micronaut.http.client.exceptions.ReadTimeoutException: Read Timeout
	at io.micronaut.http.client.exceptions.ReadTimeoutException.<clinit>(ReadTimeoutException.java:26)
	at io.micronaut.http.client.netty.DefaultHttpClient.lambda$exchangeImpl$28(DefaultHttpClient.java:1156)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:94)
	at reactor.core.publisher.SerializedSubscriber.onError(SerializedSubscriber.java:124)
	at reactor.core.publisher.FluxTimeout$TimeoutMainSubscriber.handleTimeout(FluxTimeout.java:296)
	at reactor.core.publisher.FluxTimeout$TimeoutMainSubscriber.doTimeout(FluxTimeout.java:281)
	at reactor.core.publisher.FluxTimeout$TimeoutTimeoutSubscriber.onNext(FluxTimeout.java:420)
	at reactor.core.publisher.FluxOnErrorReturn$ReturnSubscriber.onNext(FluxOnErrorReturn.java:162)
	at reactor.core.publisher.MonoDelay$MonoDelayRunnable.propagateDelay(MonoDelay.java:270)
	at reactor.core.publisher.MonoDelay$MonoDelayRunnable.run(MonoDelay.java:285)
	at io.micronaut.core.propagation.PropagatedContext.lambda$wrap$3(PropagatedContext.java:211)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:68)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:28)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
[...]

PR #1641, opened 2024-03-15
Offload JWK Set fetching from the event loop

jeremyg484:
`JwtTokenValidator` is modified to explicitly execute `JwtValidator.validate` in a separate thread from the Netty event loop.

`JwtValidator.validate` is blocking in its current implementation and this allows it to be executed in a non-blocking fashion.

This is a short-term fix for #1633, where the preferred long-term fix would be to refactor the `JwtValidator` API to be non-blocking.

NOTE: For this fix, I have chosen to simply use the standard Reactor Bounded Elastic scheduler for the separate thread execution, rather than requiring modification of the `JwtTokenValidator` constructor to inject an ExecutorService. Refactoring `JwtValidator` and `JwksSignature` to be fully non-blocking would remove the need for the explicit scheduler, thus my preference for not changing the constructor.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidator.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidator.java
@@ -22,9 +22,12 @@ import io.micronaut.security.token.jwt.signature.SignatureConfiguration;
 import io.micronaut.security.token.validator.TokenValidator;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
-import java.util.Collection;
 import org.reactivestreams.Publisher;
-import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+import java.util.Collection;
 
 /**
  * @see <a href="https://connect2id.com/products/nimbus-jose-jwt/examples/validating-jwt-access-tokens">Validating JWT Access Tokens</a>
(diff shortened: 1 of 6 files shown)


security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidator.java
```diff
@@ -67,6 +71,7 @@ public JwtTokenValidator(JwtValidator<T> validator,
                              JwtAuthenticationFactory jwtAuthenticationFactory) {
         this.validator = validator;
         this.jwtAuthenticationFactory = jwtAuthenticationFactory;
+        this.scheduler = Schedulers.boundedElastic();
```
sdelamo, line 74: Would it be better to inject an executor service instead: 

```java
@Named(TaskExecutors.BLOCKING) ExecutorService executorService) { 
...
        this.scheduler = Schedulers.fromExecutorService(executorService);
```

jeremyg484: I spent probably too much time debating this exact thing with myself. :) As alluded to in the PR description, I feel like the better long-term fix would be to refactor `JwtValidator` and on down the chain through `JwksSignature` to be non-blocking, returning Publisher instead of Optional, etc. With that in mind, my intention was to avoid having to deprecate this constructor in favor of one with the injected Executor since a non-blocking version of the API wouldn't require it. 

I suppose instead of having to deprecate the existing one(s) now, we could add a version with the Executor parameter as an option and mark that one with @Inject. But then we'd probably end up marking that version as deprecated later once a non-blocking API were in place.

sdelamo (approved):

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JwtTokenValidator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1641/commits/2d883750166a49493583365871af568b0aa517ff

PR #1660, opened 2024-04-04
Remove blocking optional

sdelamo:
This PR works around [#10609](https://github.com/micronaut-projects/micronaut-core/pull/10609) restrictions. 

This PR is not ideal as essentially sleeps and retries. I think we need to rethink the APIs and have a `ReactiveSignatureConfiguration` and reactive token validation APIs.  

I tried [`subscribesOn`](https://github.com/micronaut-projects/micronaut-security/pull/1661) or with [`@ExecutesOn`](https://github.com/micronaut-projects/micronaut-security/pull/1662/files) but it does not work.


yawkat:
This still blocks the event loop. It is not a workaround, it just hides the bug.

sdelamo:
@yawkat what do you think about https://github.com/micronaut-projects/micronaut-security/pull/1663 ?

PR #1663, opened 2024-04-04
Keep blocking with executor service

sdelamo:
This PR works around [#10609](https://github.com/micronaut-projects/micronaut-core/pull/10609) restrictions. 

The idea is to run the blocking code inside `ExecutorService::submit` of the blocking executor service.  

I tried [`subscribesOn`](https://github.com/micronaut-projects/micronaut-security/pull/1661) or with [`@ExecutesOn`](https://github.com/micronaut-projects/micronaut-security/pull/1662/files) but they do not work. 

Close: #1662 #1661 #1660


yawkat:
@sdelamo ive pushed a change that fixes the publisher runner

sdelamo:
> @sdelamo ive pushed a change that fixes the publisher runner

I added back https://github.com/micronaut-projects/micronaut-security/pull/1663/commits/ba1ea075d47b7d51c97d3f999d162063fff8c5a7 which I think you removed by mistake.

yawkat:
i removed it because the tests passed without it. not sure if it's necessary but it cant hurt.

sdelamo:
> i removed it because the tests passed without it. not sure if it's necessary but it cant hurt.

the tests that fail without it are the AOT tests which you can run locally via `./test-suite-aot.sh`

PR #1665, opened 2024-04-05
breaks: delete deprecated code

sdelamo:


sdelamo:
@wetted please add the necessary accept-api changes and docs for this breaking changes.

wetted (approved):
I see there is a breaking changes document in the user guide, that just notes deprecated code is removed. Is that enough?

wetted:
> I see there is a breaking changes document in the user guide, that just notes deprecated code is removed. Is that enough?

never mind. Now I see you comment asking me to do that. :)

wetted:
ping @sdelamo I think this is now GTG

PR #1693, opened 2024-04-30
feat: Add reactive API for JWT and JWKS

sdelamo:
This PR creates the necessary reactive APIs so that we no longer have to block. We were missing reactive APIs to fetch Signature configurations which we need for remote Json Web Key Set. 

Thus, we can remove the following workarounds: 

- https://github.com/micronaut-projects/micronaut-security/pull/1663
- https://github.com/micronaut-projects/micronaut-security/pull/1641

This PR deprecates and no longer uses the following classes: 

- `io.micronaut.security.token.jwt.validator.JwtTokenValidator`
- `io.micronaut.security.token.jwt.validator.JwtValidator`
- `io.micronaut.security.oauth2.endpoint.token.request. password.OauthPasswordAuthenticationProvider`
- `io.micronaut.security.oauth2.endpoint.token.response.validation.OpenIdTokenResponseValidator.java`

When possible, the new API uses generic types to enable the usage of libraries other than Nimbus JOSE+JWT.

Creates the following APIs: 

- `ReactiveSignatureConfiguration` - Reactive Configuration for signature verification.
- `JsonWebTokenSignatureValidator` - JSON Web Token signature validation.
- `ReactiveJsonWebTokenSignatureValidator` - JSON Web Token signature validation.
- `JsonWebTokenValidator` - JSON Web Token Validation.
- `ReactiveJsonWebTokenSignatureValidator` - Reactive JSON Web Token Validation.
- ` JsonWebTokenEncryption` - Decrypts a JSON Web Token.
- `JsonWebTokenParser`. Given a String returns a JSON Web Token. 
- `ReactiveOpenIdTokenResponseValidator` - Reactive validation of an OpenID token Response.
- `ReactiveOauthPasswordAuthenticationProvider` - Reactive   authentication provider that delegates to an OAuth 2.0 provider using the password grant flow.
- `ReactiveOpenIdPasswordAuthenticationProvider` - Reactive  authentication provider that delegates to an OpenID provider using the password grant flow.

It places the Nimbus JOSE+JWT implementations of these new APIs in the package `io.micronaut.security.token.jwt.nimbus`. It uses [`@Configuration`](https://docs.micronaut.io/4.4.6/guide/#beanConfigurations) with `@Requires(classes = JWT.class)`. 

This PR also adds a JWT TCK. This paves the way for another implementation of Micronaut Security JWT with [JJWT](https://github.com/jwtk/jjwt).

PR #1694, opened 2024-05-02
Reactive cleanup

jeremyg484:
Some minor cleanup of Reactor API usage in #1693.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/password/ReactiveOpenIdPasswordAuthenticationProvider.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/password/ReactiveOpenIdPasswordAuthenticationProvider.java
@@ -31,6 +31,7 @@ import io.micronaut.security.oauth2.endpoint.token.response.OpenIdAuthentication
 import io.micronaut.security.oauth2.endpoint.token.response.OpenIdClaims;
 import io.micronaut.security.oauth2.endpoint.token.response.validation.ReactiveOpenIdTokenResponseValidator;
 import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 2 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/request/password/ReactiveOpenIdPasswordAuthenticationProvider.java
```diff
@@ -79,19 +80,19 @@ public ReactiveOpenIdPasswordAuthenticationProvider(OauthClientConfiguration cli
     @Override
     public Publisher<AuthenticationResponse> authenticate(T requestContext, AuthenticationRequest<I, S> authenticationRequest) {
         OpenIdPasswordTokenRequestContext openIdPasswordTokenRequestContext = new OpenIdPasswordTokenRequestContext(authenticationRequest, secureEndpoint, clientConfiguration);
-        return Mono.from(tokenEndpointClient.sendRequest(openIdPasswordTokenRequestContext))
```
sdelamo, line 82: why is better to use Flux instead of Mono?. Since the reactive sequence will emit at most one item. Is it not more correct to use Mono instead?

jeremyg484: When using `Publisher` as the return value and needing to chain `flatMap` operators together, I find it more natural to use `Flux` because its `flatMap` operator takes a `Publisher`, whereas the `flatMap` operator of `Mono` requires another `Mono`, thus an additional call to `Mono.from`, introducing another layer of wrapping that makes the call stack deeper. Plus when returning `Publisher`, the semantic value of using `Mono` is essentially lost.

If we weren't returning `Publisher`, then I would indeed prefer `Mono` in this case.

sdelamo (approved):

Issue #1698, opened 2024-05-06
Cusom Secured-based annotations don't work


dantonyuk:
### Expected Behavior

When creating a new annotation that uses `@Secured` as a stereotype, it is expected to work the same way the `@Secured` annotation does. 

### Actual Behaviour

The AuthenticationException is thrown.

The reason for that is that `SecuredAnnotationRule` uses two inconsistent functions: https://github.com/micronaut-projects/micronaut-security/blob/fe91a1bc567aa916d5df85798c8535b62d9dd960/security/src/main/java/io/micronaut/security/rules/SecuredAnnotationRule.java#L73-L74

1. `hasAnnotation` that does not take stereotypes into account
2. `getAnnotation` that does.

The actual fix would just use `getAnnotation` and check if the result is `null`:

```diff
-            if (methodRoute.hasAnnotation(Secured.class)) {
-                AnnotationValue<Secured> securedAnnotation = methodRoute.getAnnotation(Secured.class);
+            AnnotationValue<Secured> securedAnnotation = methodRoute.getAnnotation(Secured.class);
+            if (securedAnnotation != null) {
```

### Steps To Reproduce

E.g., we introduced a custom annotation:

```kotlin
package com.favordelivery.micronaut.auth.token.annotations

import io.micronaut.security.annotation.Secured
import io.micronaut.security.rules.SecurityRule
import java.lang.annotation.Inherited

@Target(AnnotationTarget.FUNCTION, AnnotationTarget.TYPE)
@Retention(AnnotationRetention.RUNTIME)
@Inherited
@MustBeDocumented
@Secured(SecurityRule.IS_AUTHENTICATED)
annotation class CustomSecured(

    // some specific attributes
)
```

and try to use it in a controller handler:

```kotlin
@CustomSecured(...)
fun coolEndpoint() {
  ...
}
```
The expected behavior is for the authentication part to work as if `@Secured` annotation was placed on the handler along with the `@CustomSecured` annotation.

### Version

4.x

graemerocher:
seems like you have diagnosed the issue, perhaps you can send a simple PR with a test to use `hasStereotype`?

dantonyuk:
Thank you @graemerocher!

The PR is sent, I don't think we need to use `hasStereotype`, checking on `null` should be enough.

dantonyuk:
Closed by merging #1699.

graemerocher:
Thanks for the contribution!

PR #1699, opened 2024-05-06
fix: allow annotations stereotyped by @Secured (#1698)

dantonyuk:
See #1698 for more details.


sdelamo (approved):
Thanks for the contribution

Issue #1774, opened 2024-08-22
Violation of OpenID Connect discovery spec breaks OIDC login flow


meredrica:
### Expected Behavior

OIDC connect flow should follow the official specs and not crash with additional `token_endpoint_auth_methods_supported` types.

### Actual Behaviour

When starting an OIDC flow with a provider that sends values that are not listed in the [AuthenticationMethod](https://github.com/micronaut-projects/micronaut-security/blob/4.10.x/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/AuthenticationMethod.java) enum, the value cannot be parsed to an enum constant and thus crashes.

The OpenID Connect Discovery Spec explicitly allows sending arbitrary additional methods in `token_enpoint_auth_methods_supported`, see here: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata (...  Other authentication methods **MAY be defined by extensions**...)

Spring faced the same problem, here is the related issue: https://github.com/spring-projects/spring-security/issues/4313

The issue stems from the code in `OpenIdProviderMetadata` that parses the authentication method here: [OpenIdProviderMetadata#getTokenEndpointAuthMethods](https://github.com/micronaut-projects/micronaut-security/blob/4.10.x/security-oauth2/src/main/java/io/micronaut/security/oauth2/client/OpenIdProviderMetadata.java#L375)

It uses the `AuthenticationMethod` class, which links to the [OpenId Connect Core>Client Authentication](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) and to [Mutual TLS Profiles for OAuth Clients](https://datatracker.ietf.org/doc/html/draft-campbell-oauth-mtls) and, as far as I can tell, implements the union of all authentications mentioned there. Note that the [Mutual TLS auth for Oauth 2.0](https://datatracker.ietf.org/doc/html/rfc8705#section-2.2.1) RFC also contains the (in this case) offending `self_signed_tls_client_auth` method.

All the different (and somewhat confusing) RFCs and specs aside, it's valid to specify arbitrary values in `token_endpoint_auth_methods_supported` under the OpenID connect discovery spec, but this is currently not supported.

A full stack is attached, I think the interesting parts are these:
```
ERROR i.m.http.server.RouteExecutor - Unexpected error occurred: No enum constant io.micronaut.security.oauth2.endpoint.AuthenticationMethod.SELF_SIGNED_TLS_CLIENT_AUTH
[...]

PR #1790, opened 2024-09-06
AuthenticationMethod as a String instead of enum to support custom auth methods 

sdelamo:
See: https://github.com/micronaut-projects/micronaut-security/issues/1774

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
@@ -17,6 +17,8 @@ package io.micronaut.security.oauth2.endpoint;
 
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
+
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -29,14 +31,14 @@ import java.util.Optional;
 public class DefaultSecureEndpoint implements SecureEndpoint {
 
     private final String url;
-    private final List<AuthenticationMethod> supportedAuthenticationMethods;
+    private final List<String> supportedAuthenticationMethods;
 
     /**
      * @param url The endpoint URL
      * @param supportedAuthenticationMethods The endpoint authentication methods
      */
     public DefaultSecureEndpoint(@NonNull String url,
-                                 @Nullable List<AuthenticationMethod> supportedAuthenticationMethods) {
+                                 @Nullable List<String> supportedAuthenticationMethods) {
         this.url = url;
         this.supportedAuthenticationMethods = supportedAuthenticationMethods;
     }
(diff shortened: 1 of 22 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
```diff
@@ -29,14 +31,14 @@
...
 
     private final String url;
-    private final List<AuthenticationMethod> supportedAuthenticationMethods;
+    private final List<String> supportedAuthenticationMethods;
 
     /**
      * @param url The endpoint URL
      * @param supportedAuthenticationMethods The endpoint authentication methods
      */
     public DefaultSecureEndpoint(@NonNull String url,
```
dstepanov, line 53: Should we keep the previous constructor or is it internal class?

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
```diff
@@ -48,7 +50,28 @@ public String getUrl() {
     }
 
     @Override
-    public Optional<List<AuthenticationMethod>> getSupportedAuthenticationMethods() {
+    public Optional<List<String>> getAuthenticationMethodsSupported() {
```
dstepanov: It might be better to have `Set` if the collection has unique items and `contains` is used

dstepanov: Does it need to be optional?

security-oauth2/src/test/groovy/io/micronaut/security/oauth2/DefaultSecureEndpointTest.java
```diff
@@ -0,0 +1,43 @@
...
+class DefaultSecureEndpointTest {
+    private static final List<String> METHODS = List.of("client_secret_basic",
+            "client_secret_post",
+            "client_secret_jwt",
+            "private_key_jwt",
+            "tls_client_auth",
+            "self_signed_tls_client_auth");
+
+    @Test
+    void testGetAuthenticationMethodsSupported() {
```
dstepanov, line 26: I'm not a big fan of mock tests. Can this be easily tested using the bean context?

dstepanov (changes requested):

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
sdelamo: We cannot. Java does not allow both constructors. 

```java
public DefaultSecureEndpoint(@NonNull String url,
                                 @Nullable List<AuthenticationMethod> supportedAuthenticationMethods) {
public DefaultSecureEndpoint(@NonNull String url,
                                 @Nullable List<String> supportedAuthenticationMethods) {
```
We get a compilation method complaining about both methods have same the erasure.

sdelamo: Since, I changed to` Set<String>` I was able to keep the old constructor.

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/DefaultSecureEndpoint.java
sdelamo: I [changed to `Set<String>`](https://github.com/micronaut-projects/micronaut-security/pull/1790/commits/5685b551d623b33e86974e928fa69ae4fe35f40c) and I removed the [optional from the return type](https://github.com/micronaut-projects/micronaut-security/pull/1790/commits/574ec725738ee427b855a3c685702d4e01314ab4).

security-oauth2/src/test/groovy/io/micronaut/security/oauth2/DefaultSecureEndpointTest.java
sdelamo: `DefaultSecureEndpoint` is not a bean. I added this test to fix sonar coverage notice. I am looking if I can add an extra test using the context.

dstepanov (approved):

PR #1815, opened 2024-09-27
Fix: Implement a reactive cache for the JWKS fetching

### LINKED ISSUE #1734: Reative implementation of validation with remote JWKS fetch keys on every validations
### Expected Behavior

Remote JWKS validation should be cached and fetched only when expired as defined by JwksSignatureConfiguration::getCacheExpiration

### Actual Behaviour

Every time a JWT needs to be validated, the remote JWKS is fetched

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.5.0

timsearle:
Closes #1734 

As described on #1734, when the reactive JWKS fetching was added, the caching functionality was removed.

This pull request introduces a reactive `.cache()` alongside some expiry logic to determine when to invalidate the publisher.

### Separate Note

@sdelamo - I think there is an opportunity for a design discussion for future changes, as I believe the current approach has some shortcomings. I did begin implementing some changes, but felt that the complexity grew too much and kept this implementation simple. 

The shortcomings I see are:

- If you have multiple `JwksSignatureConfiguration`, each validator is enumerated until one succeeds - which could lead to multiple JWKS fetches. Rather than allowing someone to specify an `issuer` in their configuration, to quickly identify the appropriate JwksSignatureConfiguration to use from a JWT if the `iss` claim is present.
- If an Authorisation Server rotates a key set, a service will potentially reject newly minted JWTs until the cache expires. An idea I implemented separately was to invalidate the cache if a `kid` is observed that is not within the cache - there are trade-offs to this also worth discussion.

I have rough implementations for the first option, and the second option, but the first did include a change to the configuration to include an `issuer`, so I think it's worth discussing.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/nimbus/ReactiveJwksSignature.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/nimbus/ReactiveJwksSignature.java
@@ -104,12 +104,13 @@ public class ReactiveJwksSignature implements ReactiveSignatureConfiguration<Sig
     }
 
     private Mono<JWKSet> fetchJwkSet() {
-        return Mono.from(jwkSetFetcher.fetch(jwksSignatureConfiguration.getName(), jwksSignatureConfiguration.getUrl()));
+        return Mono.from(jwkSetFetcher.fetch(jwksSignatureConfiguration.getName(), jwksSignatureConfiguration.getUrl()))
+            .switchIfEmpty(Mono.just(new JWKSet()));
     }
 
     private record JwksCacheEntry(JWKSet jwkSet, Instant cacheExpiryAt) {
         private boolean isExpired() {
-            return cacheExpiryAt != null && Instant.now().isAfter(cacheExpiryAt);
+            return jwkSet.isEmpty() || (cacheExpiryAt != null && Instant.now().isAfter(cacheExpiryAt));
         }
     }
 }
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/nimbus/ReactiveJwksSignature.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/nimbus/ReactiveJwksSignature.java
@@ -40,11 +40,6 @@ import reactor.core.publisher.Mono;
 */
 @EachBean(JwksSignatureConfiguration.class)
 public class ReactiveJwksSignature implements ReactiveSignatureConfiguration<SignedJWT> {
-    private record JwksCacheEntry(JWKSet jwkSet, Instant cacheExpiryAt) {
-        private boolean isExpired() {
-            return cacheExpiryAt != null && Instant.now().isAfter(cacheExpiryAt);
-        }
-    }
 
     private static final Logger LOG = LoggerFactory.getLogger(ReactiveJwksSignature.class);
     private final JwkValidator jwkValidator;
@@ -110,5 +105,11 @@ public class ReactiveJwksSignature implements ReactiveSignatureConfiguration<Sig
 
     private Mono<JWKSet> fetchJwkSet() {
         return Mono.from(jwkSetFetcher.fetch(jwksSignatureConfiguration.getName(), jwksSignatureConfiguration.getUrl()));
+    }
+
+    private record JwksCacheEntry(JWKSet jwkSet, Instant cacheExpiryAt) {
+        private boolean isExpired() {
+            return cacheExpiryAt != null && Instant.now().isAfter(cacheExpiryAt);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


sdelamo:
Thanks @timsearle let me think about this and let's fix this for 4.7.0.

timsearle:
Thanks @sdelamo - ~~is there any action on me right now?~~

~~Noticed that `NotAvailableRemoteJwksSpec` is failing - as it seems the expected behaviour is we reject a JWT if the JWKS cannot be loaded, and should not propagate an error. Will investigate.~~

Resolved ✅

timsearle:
Any news/feedback on this @sdelamo? 

Currently running this via a monkey patch in production, with regular key rotations, and seems to perform well.

sdelamo:
@timsearle what do you think https://github.com/micronaut-projects/micronaut-security/pull/1824 ? This way we could offload the caching to Micronaut Cache. That would more robust, less logic to maintain in Micronaut Security and give more flexibility to users.

timsearle:
@sdelamo Will take a look in detail shortly - my immediate thought is that this retains the _changed_ behaviour in a minor version bump of Micronaut Security. 

If the behaviour is becoming opt-in, feels like a major version change? The performance profile of an API is impacted considerably.

This was quite a regression for myself when we upgraded Micronaut Security.

There is also still the limitation that you cannot have a high cache lifespan in the current solution as there is no cache-invalidation if a new `kid` is seen.

sdelamo:
> If the behaviour is becoming opt-in, feels like a major version change? The performance profile of an API is impacted considerably.

I was thinking of modifying [starter](https://launch.micronaut.io) and when you add `micronaut-security-jwt`, add a cache implementation and the necessary configuration to the config. That it is to say, new applications will have JWKS cache preconfigured. 

I don't want to have to maintain cache functionality in security when we have a module dedicated to caching.  That should have been the approach I should have followed since the beginning. 

> There is also still the limitation that you cannot have a high cache lifespan in the current solution as there is no cache-invalidation if a new kid is seen.

I was thinking to recommend an expiration of 24s after write. You will pay the penalty of retrieving a JWKS every 24 hour. since keys rotate with a higher fequency you will be fine and never had an invalid configuration.

timsearle:
> I don't want to have to maintain cache functionality in security when we have a module dedicated to caching. That should have been the approach I should have followed since the beginning.

I think this is a fair debate - but it is a major change to the behaviour of existing functionality. Can we discuss and apply the lesson in 5.x? It is a regression (the test was disabled).

> I was thinking to recommend an expiration of 24s after write. You will pay the penalty of retrieving a JWKS every 24 hour. since keys rotate with a higher fequency you will be fine and never had an invalid configuration.

I currently have mine left to the default of 60 seconds and it seems to be sufficient.

graemerocher:
I agree if there was a regression in behaviour this behaviour should be reinstated. Making it more easily configurable sounds like a 5.x feature

sdelamo:
@timsearle I have added your approach as default in https://github.com/micronaut-projects/micronaut-security/pull/1824. Please leave any comments in that PR.

timsearle:
Thanks @sdelamo @graemerocher. Left a small minor/nit comment.

For future, would appreciate some more communication and transparency as it hasn't been a great open-source contribution experience. The regression had been open since June with no indication it would be worked on/picked up, and no feedback/discussion was shared on this PR, that was intentionally very small in scope. 

Now there is a new PR, with wider scoped changes, reviewed/approved in under 24 hours, taking from this one. I was open to making changes/collaborate.

sdelamo:
> Thanks @sdelamo @graemerocher. Left a small minor/nit comment.
> 
> For future, would appreciate some more communication and transparency as it hasn't been a great open-source contribution experience. The regression had been open since June with no indication it would be worked on/picked up, and no feedback/discussion was shared on this PR, that was intentionally very small in scope.
> 
> Now there is a new PR, with wider scoped changes, reviewed/approved in under 24 hours, taking from this one. I was open to making changes/collaborate.

Thanks, good feedback. I will try to have a more open/communicative approach in the future.

timsearle:
> > Thanks @sdelamo @graemerocher. Left a small minor/nit comment.
> > For future, would appreciate some more communication and transparency as it hasn't been a great open-source contribution experience. The regression had been open since June with no indication it would be worked on/picked up, and no feedback/discussion was shared on this PR, that was intentionally very small in scope.
> > Now there is a new PR, with wider scoped changes, reviewed/approved in under 24 hours, taking from this one. I was open to making changes/collaborate.
> 
> Thanks, good feedback. I will try to have a more open/communicative approach in the future.

It appears that https://github.com/micronaut-projects/micronaut-security/pull/1824 introduced the following regression https://github.com/micronaut-projects/micronaut-security/issues/1881, because the minor upgrade had a breaking a change if someone did not introduce the new configuration property.

PR #1826, opened 2024-10-18
CSRF (Cross Site Request Forgery) Protection

### LINKED ISSUE #1074: Consider providing an API to ease the implementation of CSRF Prevention
[Cross Site Request Forgery](https://owasp.org/www-community/attacks/csrf)

sdelamo:
After this PR gets merged, I will do a PR in Views to add a ViewModelProcessor which adds the CSRF token to the views seamlessly. I will also explore add it to the form generation api.


security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    @Singleton
+    public Publisher<String> resolveToken(HttpRequest<?> request) {
+        if (request instanceof ServerHttpRequest<?> serverHttpRequest) {
+            return resolveToken(serverHttpRequest);
+        }
+        return Publishers.empty();
+    }
+
+    private Publisher<String> resolveToken(ServerHttpRequest<?> request) {
```
sdelamo: @yawkat is this the idiomatic way to get a form-url-encoded body field ?

yawkat: can't use toByteArrayPublisher because that may split form fields in the middle. have to buffer all of it. i changed it.

i wouldn't call this the idiomatic way, there simply is no idiomatic way to access form data in a server-independent way yet. but i have no better idea for now than this.

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,192 @@
...
+ * @author Sergio del Amo
+ * @since 4.11.0
+ */
+@Internal
+@Requires(property = CsrfFilterConfigurationProperties.PREFIX + ".enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
+@Requires(classes = { ExceptionHandler.class, HttpRequest.class })
+@Requires(beans = { CsrfTokenValidator.class })
+@Filter(patternStyle = FilterPatternStyle.REGEX,
+        value = "${" + CsrfFilterConfigurationProperties.PREFIX + ".regex-pattern:" + CsrfFilterConfigurationProperties.DEFAULT_REGEX_PATTERN + "}")
+final class CsrfFilter implements Ordered, HttpServerFilter {
```
yawkat: I think you can make this a method-based filter that returns a Publisher<Optional<HttpResponse<?>>>, with an empty optional if the request should pass. It should be much faster.

sdelamo: I originally tried with a method-based filter but I was not able to work with `Publisher<HttpResponse>`and I [changed it to a HttpServerFilter](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/63a6c6be312947699037d1f363ade76d0770b938).

security-csrf/src/main/java/io/micronaut/security/csrf/CsrfConfigurationProperties.java
```diff
@@ -0,0 +1,265 @@
...
+import io.micronaut.security.config.SecurityConfigurationProperties;
+import io.micronaut.security.token.generator.AccessTokenConfigurationProperties;
+
+import java.time.Duration;
+import java.time.temporal.TemporalAmount;
+import java.util.Optional;
+
+@Internal
+@ConfigurationProperties(CsrfConfigurationProperties.PREFIX)
+class CsrfConfigurationProperties implements CsrfConfiguration {
```
graemerocher: if it is internal it can be `final`

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
graemerocher: would be good to know the issues with method based filters and resolve them

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,192 @@
...
+                : ReactiveCsrfTokenResolver.of(csrfTokenResolvers, reactiveCsrfTokenResolvers);
+        this.csrfTokenValidator = csrfTokenValidator;
+        this.exceptionHandler = exceptionHandler;
+        this.csrfFilterConfiguration = csrfFilterConfiguration;
+    }
+
+    @Override
+    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
+        Supplier<Publisher<MutableHttpResponse<?>>> proceedRequest = () -> chain.proceed(request);
+        Supplier<Publisher<MutableHttpResponse<?>>> denyRequest = () -> Mono.just(unauthorized(request));
```
graemerocher: move this further down to avoid the allocation

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: I changed it https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/00df7d64cc7c9a8583c6d091d3cb785634922234

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,192 @@
...
+                            LOG.debug("CSRF Token resolved");
+                            if (csrfTokenValidator.validateCsrfToken(request, csrfToken)) {
+                                return true;
+                            } else {
+                                LOG.debug("CSRF Token validation failed");
+                                return false;
+                            }
+                        }))
+                .next()
+                .flatMap(validToken -> Mono.from(proceedRequest.get()))
```
graemerocher: don't understand why the `from` call is needed here, flatMap already accepts a publisher return type

security-csrf/src/main/java/io/micronaut/security/csrf/repository/CompositeCsrfTokenRepository.java
```diff
@@ -0,0 +1,50 @@
...
+    public CompositeCsrfTokenRepository(List<CsrfTokenRepository<T>> repositories) {
+        this.repositories = repositories;
+    }
+
+    @Override
+    public Optional<String> findCsrfToken(T request) {
+        return repositories.stream()
+                .map(r -> r.findCsrfToken(request))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
graemerocher: these 3 calls can be replaced by `.flatMap(r -> r.findCsrfToken(request).stream())`

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
```diff
@@ -0,0 +1,52 @@
...
+        this.sessionIdResolvers = sessionIdResolvers;
+    }
+
+    @Override
+    @NonNull
+    public Optional<String> findSessionId(@NonNull T request) {
+        return sessionIdResolvers.stream()
+                .map(sessionIdResolver -> sessionIdResolver.findSessionId(request))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
graemerocher: again these 3 calls can be replaced by the `stream()` method

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
graemerocher: I think we should add a way to resolve the form data to map in a server-independent way for netty I guess it can be done by splitting the body and buffering.

Not sure how to handle multiple part requests however.

graemerocher (changes requested):

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: If I replace it to: 

```java
 .flatMap(validToken -> proceedRequest.get())
```

it complains with: 

```
no instance(s) of type variable(s) R exist so that Publisher<Optional<MutableHttpResponse<?>>> conforms to Mono<? extends R>
```

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/1effdbf461849e9381a73318c54ca7eea99e3330

security-csrf/src/main/java/io/micronaut/security/csrf/repository/CompositeCsrfTokenRepository.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/50998738265a2159b2c38c7e4f161ca98a967497

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/77a157906008ec100b846761b90b94e7186368ff

security-csrf/src/main/java/io/micronaut/security/csrf/CsrfConfigurationProperties.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/0fa439125b76dadad0c2adc113d5dcc722cac937

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
sdelamo: I am going to try to create a PR in core. Currently, `ServerHttpRequest` is netty only.

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
```diff
@@ -0,0 +1,80 @@
...
+ * @param <T> Request
+ * @since 4.11.0
+ * @author Sergio del Amo
+ */
+@Requires(beans = { CsrfTokenRepository.class, DefaultCsrfTokenGenerator.class})
+@Singleton
+public class RepositoryCsrfTokenValidator<T> implements CsrfTokenValidator<T> {
+    private static final Logger LOG = LoggerFactory.getLogger(RepositoryCsrfTokenValidator.class);
+    private final List<CsrfTokenRepository<T>> repositories;
+    private final DefaultCsrfTokenGenerator<T> defaultCsrfTokenGenerator;
```
n0tl3ss: Can `DefaultCsrfTokenGenerator` be replaced with`CsrfTokenGenerator`? I mean the implementation of interfaces might be extended that you can provide randomValue that will be used if provided. In that case when you provide one from the session you should get same with hmac. The `RepositoryCsrfTokenValidator` looks can be more generic and used with some other implementations?

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
yawkat: @graemerocher i plan to make a common API for this eventually, but it will be much more complicated because it will support streaming individual parts of the body independently

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
yawkat: you may have to give explicit type parameters, something like `.<HttpResponse<?>>flatMap(...)`

yawkat:
I haven't reviewed in detail yet, will do that later.

n0tl3ss (approved):

n0tl3ss:
Btw @sdelamo great job! :)

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
sdelamo: see: https://github.com/micronaut-projects/micronaut-core/pull/11272

yawkat: @sdelamo also, ServerHttpRequest is not netty only. servlet also implements it, and so do the cloud modules

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
sdelamo: To replace `CsrfTokenGenerator`, we will need expose the hmac method in the `CsrfTokenGenerator`. Not sure, we want to do that.

n0tl3ss: the thing is you might not have to do. the `CsrfTokenGenerator` generateCsrfToken token method can be like
```
    @NonNull
    String generateCsrfToken(@NonNull T request, @Nullable String randomValue);
```

sdelamo: I would prefer to expose the `hmac` method in `CsrfTokenGenerator`. I expect, users to use `CsrfTokenGenerator` directly and I would prefer if they don't need to think about the random value generation.

src/main/docs/guide/csrf/csrfApis.adoc
```diff
@@ -0,0 +1,9 @@
+The main APIs for CSRF protection are:
+
+* api:security.csrf.CsrConfiguration[]
+* api:security.csrf.filter.CsrFilter[]
+* api:security.csrf.filter.CsrFilterConfiguration[]
+* api:security.csrf.resolvers.CsrfTokenResolver[]
+* api:security.csrf.generator.CsrfTokenGenerator[]
+* api:security.csrf.CsrTokenValidator[]
```
yawkat: Inconsistent naming? Also I think the package name is wrong

src/main/docs/guide/csrf/csrfMitigations/doubleSubmitCookiePattern.adoc
```diff
@@ -0,0 +1,21 @@
...
+Cookie: session=<token>; __Host-csrfToken=o24b65486f506e2cd4403caf0d640024
+[...]
+
+amount=100&toUser=intended&csrfToken=o24b65486f506e2cd4403caf0d640024
+----
+
+When you use Micronaut Security Authentication <<micronautSecurityAuthenticationCookie, `cookie`>>, or
+    <<micronautSecurityAuthenticationIdToken, `idtoken`>> a CSRF Token is saved in a Cookie upon login.
+
+You can <<csrfConfiguration, configure different cookie attributes>>. For example, by default the cookie name uses a `__Host-`  https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie_prefixes[Cookie prefix], can extend https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#using-cookies-with-host-prefixes-to-identify-origins[security protections against CSF Attacks].
```
yawkat:
```suggestion
You can <<csrfConfiguration, configure different cookie attributes>>. For example, by default the cookie name uses a `__Host-`  https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie_prefixes[Cookie prefix], can extend https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#using-cookies-with-host-prefixes-to-identify-origins[security protections against CSRF Attacks].
```

src/main/docs/guide/csrf/csrfTokenResolvers/fieldCsrfTokenResolver.adoc
```diff
@@ -0,0 +1,15 @@
...
+Host: vulnerable bank
+Content-Type: application/x-www-form-urlencoded
+Cookie: session=<token>
+[...]
+amount=100&toUser=intended&csrfToken=o24b65486f506e2cd4403caf0d640024
+----
+
+You can disable it by setting `micronaut.security.csrf.token-resolvers.field.enabled=false`
+
+NOTE: `FieldCsrfTokenResolver` only works for Netty runtime.
```
yawkat: should also work elsewhere

security/src/main/java/io/micronaut/security/utils/HMacUtils.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param algorithm HMAC algorithm
+     * @param data Data
+     * @param key Signature Key
+     * @return HMAC encoded in Base64
+     * @throws NoSuchAlgorithmException if no {@code Provider} supports a {@code MacSpi} implementation for the specified algorithm.
+     * @throws  InvalidKeyException if the given key is inappropriate for initializing this MAC.
+     */
+    public static String base64EncodedHmac(@NonNull String algorithm, @NonNull String data, @NonNull String key)
+            throws NoSuchAlgorithmException, InvalidKeyException {
+        SecretKeySpec secretKeySpec = new SecretKeySpec(key.getBytes(), algorithm);
```
yawkat:
```suggestion
        SecretKeySpec secretKeySpec = new SecretKeySpec(key.getBytes(StandardCharsets.UTF_8), algorithm);
```

Not a very portable way of using a configured key, but I think it doesn't need to be in this case.

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,194 @@
...
+        this.csrfTokenValidator = csrfTokenValidator;
+        this.exceptionHandler = exceptionHandler;
+        this.csrfFilterConfiguration = csrfFilterConfiguration;
+    }
+
+    @RequestFilter
+    @Nullable
+    public Publisher<Optional<MutableHttpResponse<?>>> csrfFilter(@NonNull HttpRequest<?> request) {
+        if (!shouldTheFilterProcessTheRequestAccordingToTheHttpMethod(request)) {
+            return proceedRequest();
```
yawkat: Actually it gets even better. I think you can do this:

```suggestion
            return null;
```

yawkat: Actually, hmm, not sure. Would be a nice feature if it doesn't work.

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+        if (!shouldTheFilterProcessTheRequestAccordingToTheContentType(request)) {
+            return proceedRequest();
+        }
+        return reactiveCsrfTokenResolvers.isEmpty()
+                ? imperativeFilter(request)
+                : reactiveFilter(request);
+    }
+
+    private static Publisher<Optional<MutableHttpResponse<?>>> proceedRequest() {
```
yawkat:
```suggestion
    private static Mono<Optional<MutableHttpResponse<?>>> proceedRequest() {
```

see below

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,194 @@
...
+                            LOG.debug("CSRF Token resolved");
+                            if (csrfTokenValidator.validateCsrfToken(request, csrfToken)) {
+                                return true;
+                            } else {
+                                LOG.debug("CSRF Token validation failed");
+                                return false;
+                            }
+                        }))
+                .next()
+                .flatMap(validToken -> Mono.from(proceedRequest()))
```
yawkat:
```suggestion
                .flatMap(validToken -> proceedRequest())
```

see above

security-csrf/src/main/java/io/micronaut/security/csrf/generator/DefaultCsrfTokenGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return HMAC hash
+     */
+    @NonNull
+    public String hmac(@NonNull T request, String randomValue) {
+        // Gather the values
+        String secret = csrfConfiguration.getSecretKey();
+        String sessionID = sessionIdResolver.findSessionId(request).orElse(""); // Current authenticated user session
+
+        // Create the CSRF Token
+        String message = sessionID + SESSION_RANDOM_SEPARATOR + randomValue; // HMAC message payload
```
yawkat: Is it necessary to use this format? Otherwise, to prevent attacks against the format, I suggest

```suggestion
        String message = sessionID.length() + "." + sessionID + "." + randomValue.length() + "." + randomValue; // HMAC message payload
```

yawkat: The problem is described in more detail in the first section here: https://blog.trailofbits.com/2024/08/21/yolo-is-not-a-valid-hash-construction/

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
```diff
@@ -0,0 +1,80 @@
...
+        if (arr.length != 2) {
+            if (LOG.isWarnEnabled()) {
+                LOG.warn("Invalid CSRF token: {}", csrfTokenInRequest);
+            }
+            return false;
+        }
+        String hmac = arr[0];
+        String randomValue = arr[1];
+        String expectedHmac = defaultCsrfTokenGenerator.hmac(request, randomValue);
+        return hmac.equals(expectedHmac);
```
yawkat:
```suggestion
        return MessageDigest.isEquals(expectedHmac.getBytes(StandardCharsets.UTF_8), hmac.getBytes(StandardCharsets.UTF_8));
```

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
```diff
@@ -0,0 +1,80 @@
...
+        this.defaultCsrfTokenGenerator = defaultCsrfTokenGenerator;
+    }
+
+    @Override
+    public boolean validateCsrfToken(T request, String csrfTokenInRequest) {
+        for (CsrfTokenRepository<T> repo : repositories) {
+            Optional<String> csrfTokenOptional = repo.findCsrfToken(request);
+            if (csrfTokenOptional.isPresent()) {
+                String csrfTokenInRepository = csrfTokenOptional.get();
+                if (csrfTokenInRepository.equals(csrfTokenInRequest) && validateHmac(request, csrfTokenInRequest)) {
```
yawkat, line 63: Why do we need to validate the HMAC if we already confirmed that the CSRF token matches that from the repository?

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void defaultContentType() {
+        assertEquals(Set.of(MediaType.APPLICATION_FORM_URLENCODED_TYPE, MediaType.MULTIPART_FORM_DATA_TYPE), csrfFilterConfiguration.getContentTypes());
+    }
+
+    @Test
+    void regexPatternCanBeChanged() {
+        String regexPattern = csrfFilterConfiguration.getRegexPattern();
+        assertFalse(PathMatcher.REGEX.matches(csrfFilterConfiguration.getRegexPattern(), "/login"));
+        assertTrue(PathMatcher.REGEX.matches(csrfFilterConfiguration.getRegexPattern(), "/todo/list"));
+        assertEquals("^(?!\\/login).*$", regexPattern);
```
yawkat, line 59: We need to check whether this approach is vulnerable to path traversal attacks. e.g. if the attacker triggers a request to `/login/../todo/list`

security-csrf/src/test/java/io/micronaut/security/csrf/repository/CsrfDoubleSubmitCookiePatternTest.java
```diff
@@ -0,0 +1,175 @@
...
+        assertDenied(client, cookieJwt.getValue(), csrfTokenCookieName, formWithCsrfToken, cookieCsrfToken.getValue());
+
+        // CSRF Token with HMAC but not session id feed into HMAC calculation, request is unauthorized
+        String randomValue = "abcdefg";
+        String hmac = HMacUtils.base64EncodedHmacSha256(randomValue, csrfConfiguration.getSecretKey());
+        String csrfTokenCalculatedWithoutSessionId = hmac + "." + randomValue;
+        PasswordChangeForm body = new PasswordChangeForm("sherlock", "evil", csrfTokenCalculatedWithoutSessionId);
+        assertDenied(client, cookieJwt.getValue(), csrfTokenCookieName, body, csrfTokenCalculatedWithoutSessionId);
+
+        String message = FIX_SESSION_ID + "!" + randomValue;
```
yawkat: move into a static method in the generator that is VisibleForTesting

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/IdTokenLoginHandler.java
```diff
@@ -32,17 +32,15 @@
...
 import org.slf4j.LoggerFactory;
 
 import java.text.ParseException;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
yawkat: We need a checkstyle rule for this so that we don't always have diffs changing this back and forth because of different IDE settings.

(I vote for banning star imports)

security-csrf/src/main/java/io/micronaut/security/csrf/generator/DefaultCsrfTokenGenerator.java
sdelamo: The CSRF token generation follows the [OWSAP example](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#pseudo-code-for-implementing-hmac-csrf-tokens). 

You suggested change removes the session dependent value. 

> To generate HMAC CSRF tokens (with a session-dependent user value), the system must have:
> A session-dependent value that changes with each login session. 

However, with your suggestion, different session dependent values of same length will create the same value.

What do you think about?

https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/54228f6017a62ef3e84378cb2f2179bb2cc2ea7b

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
sdelamo: is this to avoid time constant attacks?

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
sdelamo: You add the hmac generated with a secret to prevent attackers from overriding the cookie value with their own (e.g. with taken-over subdomain attacks)

src/main/docs/guide/toc.yml
```diff
@@ -33,7 +33,12 @@ endpoints:
   logout:
     title: Logout Controller
     logoutHandler: Logout Handler
-  builtInHandlers: Built-in Login and Logout Handlers
+  builtInHandlers:
+    title: Built-in Login and Logout Handlers
+    micronautSecurityAuthenticationBearer: Micronaut Security Authentication Bearer
```
graemerocher: I don't think you need to repeat `Micronaut Security` for each title. These are the Micronaut Security docs after all

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
sdelamo: that it is regular expression used to match the request's path the server filter applies. 

Not sure how an attacker will work. If they send a request to `/login/.../todo/list` they will get a 404

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/token/response/IdTokenLoginHandler.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/629b8fee4acdef3db4ac2e8694b4d74d83825a9d

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/4506c4748a809eba61ecb2ab9593f2fe702773dd

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/4506c4748a809eba61ecb2ab9593f2fe702773dd

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: It does not work to return `null. `

It does not work to return `Mono.empty()`. 

Only, it works to return `Mono.just(Optional.empty())`

That it is the reason I created:
https://github.com/micronaut-projects/micronaut-core/pull/11270

With the current javadoc, if you did not tell me that to proceed with a request in a filter method return Publisher I had to use `Publisher<Optional<HttpResponse>>` and return `Mono.just(Optional.empty())` I would not have figured it out.

src/main/docs/guide/csrf/csrfTokenResolvers/fieldCsrfTokenResolver.adoc
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/b18f32f1b48a9c8788b224e7804b8d0f50655d90

src/main/docs/guide/csrf/csrfApis.adoc
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/1f92c7c3c48866d517e91963e67cec6a054ae8f9

security-csrf/src/test/java/io/micronaut/security/csrf/repository/CsrfDoubleSubmitCookiePatternTest.java
sdelamo: I extracted the method https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/36f45cdf124125ad907cd4872c531600ed96f4da I would preferred not to extract a one line method.

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
yawkat: yep

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/4506c4748a809eba61ecb2ab9593f2fe702773dd

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
yawkat: I'm just not super confident in the security of using filter patterns for matching what is essentially routes. If the router does some shenanigans with the URI, such as "simplifying" `/login/../todo/list` to `/todo/list`, then this form of exclusion using regex may become vulnerable.

I'd be more comfortable if we could match this regex against the URI of the resolved route, because that is not attacker-controlled. Would that be possible?

security-csrf/src/test/java/io/micronaut/security/csrf/repository/CsrfDoubleSubmitCookiePatternTest.java
yawkat: ok, your call

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/07a755b4f1c445460a5c80d652134ae2c1d44f76

graemerocher (changes requested):

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
sdelamo: what API exposes the `URI of the resolved route`? 
> I'm just not super confident in the security of using filter patterns for matching what is essentially routes. If the router does some shenanigans with the URI, such as "simplifying" /login/../todo/list to /todo/list, then this form of exclusion using regex may become vulnerable.

If the router redirects a request with path `/login/../todo/list` to `/todo/list` then I think we have a vulnerability in the router. 

However, I am ok doing  an extra check in the Csrf filter. However, What API exposes the `URI of the resolved route`? I looked into `RouteMatch` but I don't see any path there.

src/main/docs/guide/toc.yml
```diff
@@ -33,7 +33,12 @@ endpoints:
...
+    micronautSecurityAuthenticationSession: Micronaut Security Session Login Handler
+    micronautSecurityAuthenticationCookie: Micronaut Security Authentication Cookie
+    micronautSecurityAuthenticationIdToken: Micronaut Security Authentication ID Token
```
sdelamo:
```suggestion
    micronautSecurityAuthenticationSession: Authentication Mode Session
    micronautSecurityAuthenticationCookie: Authentication Mode Cookie
    micronautSecurityAuthenticationIdToken: Authentication Mode ID Token
```

src/main/docs/guide/toc.yml
sdelamo:
```suggestion
    micronautSecurityAuthenticationBearer: Authentication Mode Bearer
```

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/c12365cecdbb34c42c44746224a5dd51e072fa17 and https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/3db52fa3d24c5edf2be8f44d7b62d0b7b0c344d2

security-csrf/src/main/java/io/micronaut/security/csrf/generator/DefaultCsrfTokenGenerator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/d8d5020af2080ccb7a45bcc857751164dcde65c0

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
yawkat: UriRouteMatch has a getUri method which is normalized to some extent by the router (leading slash and trailing parameters removed). Maybe that could work

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
sdelamo: I have updated the code to use the new [FilterBodyParser API](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/94beaf6f0061ab55ba53bfcab13c8f99c47fe207) and changed the [reactive API to use CompletableFuture](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/7f708037a5abd490a057f2c3b49fca63ddbe7503).

security-csrf/src/test/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationTest.java
sdelamo: something like https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/13c9cfe7898bb0d246da96be36bf2e597fbd577c ?

security-csrf/src/main/java/io/micronaut/security/csrf/CsrfConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @return Random value's size in bytes. The random value used is used to build a CSRF Token.
+     */
+    int getRandomValueSize();
+
+    /**
+     *
+     * @return The Secret Key that is used to calculate an HMAC as part of a CSRF token generation.
+     */
+    String getSecretKey();
```
graemerocher, line 43: add nullability annotation here

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private static Mono<Optional<MutableHttpResponse<?>>> proceedRequest() {
+        return Mono.just(Optional.empty());
+    }
+
+    private Mono<Optional<MutableHttpResponse<?>>> reactiveFilter(HttpRequest<?> request) {
+        return Flux.fromIterable(this.futureCsrfTokenResolvers)
+                .concatMap(resolver -> Mono.fromFuture(resolver.resolveToken(request))
+                        .filter(csrfToken -> {
+                            LOG.debug("CSRF Token resolved");
```
graemerocher: wrap these debug statements in if statements

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,223 @@
...
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Request uri {} does not match fitler regex pattern {}", uri, csrfFilterConfiguration.getRegexPattern());
+            }
+            return false;
+        }
+        return true;
+    }
+
+    private static Mono<Optional<MutableHttpResponse<?>>> proceedRequest() {
+        return Mono.just(Optional.empty());
```
graemerocher: this mono could be extracted into a constant

security-csrf/src/main/java/io/micronaut/security/csrf/generator/CsrfHmacTokenGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+    String HMAC_RANDOM_SEPARATOR = ".";
+
+    /**
+     * Generates an HMAC.
+     * @param request Request
+     * @param randomValue Cryptographic random value
+     * @return HMAC hash
+     */
+    @NonNull
+    String hmac(@NonNull T request, String randomValue);
```
graemerocher: add nullability annotation to the second argument

security-csrf/src/main/java/io/micronaut/security/csrf/repository/CsrfTokenRepository.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param <T> Request
+ */
+@FunctionalInterface
+public interface CsrfTokenRepository<T> extends Ordered {
+    /**
+     *
+     * @param request Request
+     * @return A CSRF token or an empty optional.
+     */
+    Optional<String> findCsrfToken(T request);
```
graemerocher: add nullability annotation to the argument

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/CsrfTokenResolver.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface CsrfTokenResolver<T> extends Ordered {
+
+    /**
+     *
+     * @param request The Request. Maybe an HTTP Request.
+     * @return A CSRF token or an empty Optional if the token cannot be resolved.
+     */
+    @NonNull
+    Optional<String> resolveToken(T request);
```
graemerocher: add nullability annotation to the argument

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
```diff
@@ -0,0 +1,67 @@
...
+import jakarta.inject.Singleton;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Resolves a CSRF token from a form-urlencoded body using the {@link ServerHttpRequest#byteBody()} API.
+ *
+ * @since 2.0.0
+ */
+@Requires(classes = HttpRequest.class)
+@Requires(property = "micronaut.security.csrf.token-resolvers.field.enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
```
graemerocher: extract this property into a constant

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
```diff
@@ -0,0 +1,67 @@
...
+@Singleton
+class FieldCsrfTokenResolver implements FutureCsrfTokenResolver<HttpRequest<?>> {
```
graemerocher: can this be `final`?

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+public interface FutureCsrfTokenResolver<T> extends Ordered {
+
+    /**
+     *
+     * @param request The Request. Maybe an HTTP Request.
+     * @return A CSRF token or an empty Optional if the token cannot be resolved.
+     */
+    @SingleResult
+    @NonNull
+    CompletableFuture<String> resolveToken(T request);
```
graemerocher: add nullability annotation to the argument

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     *
+     * @param resolvers Imperative CSRF Token Resolvers
+     * @param futureCsrfTokenResolvers Reactive CSRF Token Resolvers
+     * @return Returns a List of {@link FutureCsrfTokenResolver} instances containing every reactive resolver plus the imperative resolvers adapted to imperative.
+     * @param <T>
+     */
+    static <T> List<FutureCsrfTokenResolver<T>> of(List<CsrfTokenResolver<T>> resolvers,
+                                                   List<FutureCsrfTokenResolver<T>> futureCsrfTokenResolvers) {
```
graemerocher: add nullability annotations to return type and arguments

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+        List<FutureCsrfTokenResolver<T>> result  = new ArrayList<>();
```
graemerocher: size the array according to the passed arguments

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/HttpHeaderCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+        this.csrfConfiguration = csrfConfiguration;
+    }
+
+    @Override
+    public Optional<String> resolveToken(HttpRequest<?> request) {
+        String csrfToken = request.getHeaders().get(csrfConfiguration.getHeaderName());
+        if (StringUtils.isNotEmpty(csrfToken)) {
+            return Optional.of(csrfToken);
+        }
+        csrfToken = request.getHeaders().get(csrfConfiguration.getHeaderName().toLowerCase());
```
graemerocher: can this call to `toLowerCase()` be stored in a field to avoid repeated calls?

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/HttpHeaderCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final int ORDER = -100;
+    private final CsrfConfiguration csrfConfiguration;
+
+    HttpHeaderCsrfTokenResolver(CsrfConfiguration csrfConfiguration) {
+        this.csrfConfiguration = csrfConfiguration;
+    }
+
+    @Override
+    public Optional<String> resolveToken(HttpRequest<?> request) {
+        String csrfToken = request.getHeaders().get(csrfConfiguration.getHeaderName());
```
graemerocher: extract `request.getHeaders()` into a local variable to avoid repeated calls

security-csrf/src/main/java/io/micronaut/security/csrf/session/CsrfSessionPopulator.java
```diff
@@ -0,0 +1,45 @@
...
+    private final CsrfTokenGenerator<T> csrfTokenGenerator;
+
+    public CsrfSessionPopulator(CsrfConfiguration csrfConfiguration, CsrfTokenGenerator<T> csrfTokenGenerator) {
+        this.csrfConfiguration = csrfConfiguration;
+        this.csrfTokenGenerator = csrfTokenGenerator;
+    }
+
+    @Override
+    public void populateSession(T request, Authentication authentication, Session session) {
+        session.put(csrfConfiguration.getHttpSessionName(), csrfTokenGenerator.generateCsrfToken(request));
```
graemerocher, line 53: split the method calls onto multiple lines to improve readability

graemerocher (changes requested):

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/HttpHeaderCsrfTokenResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/96347417a157da5af9293e4fa9e89b631d31bad0

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/HttpHeaderCsrfTokenResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/dee17cf0e85bed034667e36c79a6a8c4dfc31253

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/0e26b76c5a4cacf4bb5f0498a35f7ddd32643ac0

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/3dd3357a7a038c9fc444dd47dced95404bd86fba

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/4cf98a24962c1acd9072fe691cf0b3c2b107f009

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/8b81613714ce6758decac40bf52d79d7d21ae033

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FieldCsrfTokenResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/828b96b2e41bcbd50d54c6e13ca91a13edeaf9cf

graemerocher (approved):

yawkat (approved):

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,224 @@
...
+                ? futureCsrfTokenResolvers
+                : FutureCsrfTokenResolver.of(csrfTokenResolvers, futureCsrfTokenResolvers);
+        this.csrfTokenValidator = csrfTokenValidator;
+        this.exceptionHandler = exceptionHandler;
+        this.csrfFilterConfiguration = csrfFilterConfiguration;
+    }
+
+    @RequestFilter
+    @Nullable
+    public Publisher<Optional<HttpResponse<?>>> csrfFilter(@NonNull HttpRequest<?> request) {
```
dstepanov: Use completable future

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+        return true;
+    }
+
+    boolean shouldTheFilterProcessTheRequestAccordingToTheUriMatch(UriRouteMatch<?, ?> uriRouteMatch) {
+        return shouldTheFilterProcessTheRequestAccordingToTheUriMatch(uriRouteMatch.getUri());
+    }
+
+    boolean shouldTheFilterProcessTheRequestAccordingToTheUriMatch(String uri) {
+        boolean matches = PathMatcher.REGEX.matches(csrfFilterConfiguration.getRegexPattern(), uri);
```
dstepanov, line 111: Compile the pattern and store in a field

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
```diff
@@ -0,0 +1,224 @@
...
+        if (!matches) {
+            if (LOG.isTraceEnabled()) {
```
dstepanov, line 113: Looks like you use a lot of this check maybe create a static field

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationProperties.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Whether the filter is enabled. Default value {@value #DEFAULT_ENABLED}.
+     * @param enabled Whether the filter is enabled.
+     */
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    @Override
+    public String getRegexPattern() {
```
dstepanov, line 104: Maybe have the property named `regex` and `regexPattern` will return an instance of `Pattern`?

security-csrf/src/main/java/io/micronaut/security/csrf/generator/CsrfHmacTokenGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+    String HMAC_RANDOM_SEPARATOR = ".";
+
+    /**
+     * Generates an HMAC.
+     * @param request Request
+     * @param base64EncodedRandomValue Cryptographic random value encoded as Base64
+     * @return HMAC hash
+     */
+    @NonNull
+    String hmac(@NonNull T request, @NonNull String base64EncodedRandomValue);
```
dstepanov, line 41: Considering you have `generateCsrfToken` maybe name the method similar

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param <T> request type
+     */
+    @NonNull
+    static <T> List<FutureCsrfTokenResolver<T>> of(
+            @NonNull List<CsrfTokenResolver<T>> resolvers,
+            @NonNull List<FutureCsrfTokenResolver<T>> futureCsrfTokenResolvers) {
+        List<FutureCsrfTokenResolver<T>> result  = new ArrayList<>(futureCsrfTokenResolvers.size() + resolvers.size());
+        result.addAll(futureCsrfTokenResolvers);
+        result.addAll(resolvers.stream().map(FutureCsrfTokenResolverAdapter::new).toList());
+        OrderUtil.sort(result);
```
dstepanov, line 58: You can use `CollectionUtils.concat`

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return Returns a List of {@link FutureCsrfTokenResolver} instances containing every reactive resolver plus the imperative resolvers adapted to imperative.
+     * @param <T> request type
+     */
+    @NonNull
+    static <T> List<FutureCsrfTokenResolver<T>> of(
+            @NonNull List<CsrfTokenResolver<T>> resolvers,
+            @NonNull List<FutureCsrfTokenResolver<T>> futureCsrfTokenResolvers) {
+        List<FutureCsrfTokenResolver<T>> result  = new ArrayList<>(futureCsrfTokenResolvers.size() + resolvers.size());
+        result.addAll(futureCsrfTokenResolvers);
+        result.addAll(resolvers.stream().map(FutureCsrfTokenResolverAdapter::new).toList());
```
dstepanov: This one I would write `.forEach(result::add)`

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolverAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import io.micronaut.core.annotation.NonNull;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Adapter from {@link CsrfTokenResolver} to {@link FutureCsrfTokenResolver}.
+ * @param <T> Request
+ */
+public class FutureCsrfTokenResolverAdapter<T> implements FutureCsrfTokenResolver<T> {
```
dstepanov: Is it a public and non-final class by design? No version too.

security-csrf/src/main/java/io/micronaut/security/csrf/session/CsrfSessionPopulator.java
```diff
@@ -0,0 +1,53 @@
...
+import jakarta.inject.Singleton;
+
+/**
+ * Populates the session with a CSRF token.
+ * @author Sergio del Amo
+ * @since 4.11.0
+ * @param <T> Request
+ */
+@Singleton
+public class CsrfSessionPopulator<T> implements SessionPopulator<T> {
```
dstepanov: Non-final and public by design?

security-csrf/src/main/java/io/micronaut/security/csrf/session/HttpSessionSessionIdResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Optional;
+
+/**
+ * Implementation of {@link SessionIdResolver} that returns {@link Session#getId()} if an HTTP session  is associated with the request.
+ * @author Sergio del Amo
+ * @since 4.11.0
+ */
+@Singleton
+public class HttpSessionSessionIdResolver implements SessionIdResolver<HttpRequest<?>> {
```
dstepanov: Non-final and public by design?

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * {@link CsrfTokenValidator} implementation that uses a {@link CsrfTokenRepository}.
+ * First attempts to retrieve a token from a {@link CsrfTokenRepository} and if found validates it against the supplied token.
+ * @param <T> Request
+ * @since 4.11.0
+ * @author Sergio del Amo
+ */
+@Requires(beans = { CsrfTokenRepository.class, CsrfHmacTokenGenerator.class})
+@Singleton
+public class RepositoryCsrfTokenValidator<T> implements CsrfTokenValidator<T> {
```
dstepanov: Public and non-final by design?

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JsonWebTokenIdSessionIdResolver.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Implementation of {@link SessionIdResolver} that returns the jti claim JWT ID if a JWT token  is associated with the request.
+ *
+ * @since 4.11.0
+ * @author  Sergio del Amo
+ */
+@Requires(classes = HttpRequest.class)
+@Requires(bean = JsonWebTokenParser.class)
+@Singleton
+public class JsonWebTokenIdSessionIdResolver implements SessionIdResolver<HttpRequest<?>> {
```
dstepanov: Non-final and public by design?

security-session/src/main/java/io/micronaut/security/session/DefaultSessionPopulator.java
```diff
@@ -0,0 +1,40 @@
...
+import io.micronaut.security.filters.SecurityFilter;
+import io.micronaut.session.Session;
+import jakarta.inject.Singleton;
+
+/**
+ * Default Implementation of {@link SessionPopulator}. It adds the {@link Authentication} object to the session with the key {@link SecurityFilter#AUTHENTICATION}.
+ * @param <T> Request
+ */
+@Singleton
+public class DefaultSessionPopulator<T> implements SessionPopulator<T> {
```
dstepanov: Non-final and public by design?

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Optional;
+
+/**
+ * Composite Pattern implementation of {@link SessionIdResolver}.
+ * @see <a href="https://guides.micronaut.io/latest/micronaut-patterns-composite.html">Composite Pattern</a>
+ * @param <T> Request
+ */
+@Primary
+@Singleton
+public class CompositeSessionIdResolver<T> implements SessionIdResolver<T> {
```
dstepanov: Non-final and public by design?
I'm not a fan of primary, can it be just a different order?

dstepanov (changes requested):

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: see. https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/aba5267a2e31ba97c5c788c966cc8356aeff7a2c

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: The log levels be changed dynamically via the logger endpoint for example. If I move the logic to a static field e.g.

```
private final static boolean isTraceEnabled = LOG.isTracedEnabled()
```

That would be set in stone and changing the log levels will not work.

security-csrf/src/main/java/io/micronaut/security/csrf/generator/CsrfHmacTokenGenerator.java
sdelamo: I think the method name fine. The method calculates a `hmac`.

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
sdelamo: [I changed the code to use `CollectionUtils.concat`](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/bb937ad8843cc61898803ff16ab502ecfcc6f05b)

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolver.java
sdelamo: [I changed the code to use `CollectionUtils.concat`](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/bb937ad8843cc61898803ff16ab502ecfcc6f05b)

security-csrf/src/main/java/io/micronaut/security/csrf/resolver/FutureCsrfTokenResolverAdapter.java
sdelamo: I made it package private and final https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/2fb98bfc6f69e57d712f67a92baa7bb0d1817838

security-csrf/src/main/java/io/micronaut/security/csrf/session/CsrfSessionPopulator.java
sdelamo: I made it final and package private https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/c70e6e6b27e893330bf0f89b7982d3c4bbc0477f

security-csrf/src/main/java/io/micronaut/security/csrf/validator/RepositoryCsrfTokenValidator.java
sdelamo: I made it package private and final https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/816466c0eaa566f1078f6682922b2cb984aaab0f

security-jwt/src/main/java/io/micronaut/security/token/jwt/validator/JsonWebTokenIdSessionIdResolver.java
sdelamo: I made it final and package private https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/37517d30649548afe82a673f58222520b289707a

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
sdelamo: [I made it final and package private](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/f2e202523ebe734213e7c6f7a8aa62eae0ba60f9)

I want it to be `@Primary` to use a composite pattern.

security-session/src/main/java/io/micronaut/security/session/DefaultSessionPopulator.java
sdelamo: I made it final and package private https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/a89d437121df06ddad0ecceb7f7f09e8fbfa149c

security-csrf/src/main/java/io/micronaut/security/csrf/session/HttpSessionSessionIdResolver.java
sdelamo: I made it package private and final https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/b6cd612e0dc43762dd6f84fb6ddda118a3eabd11

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
graemerocher: You should add `@Indexed(SessionIdResolver.class)` to the interface

security-csrf/src/main/java/io/micronaut/security/csrf/generator/CsrfHmacTokenGenerator.java
yawkat: maybe a better API would be a `verify` method. Then you wouldn't have to expose the separator character or the fact that this uses a HMAC.

sdelamo: We already have a `CsrfTokenValidator`. Users should not not use `CsrfHmacTokenGenerator` if they want to roll their own. [I have annotated it with `@Internal`.](https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/3c6632611be7c2d4c19f9ea96e3e8d3a43929f29)

security/src/main/java/io/micronaut/security/session/CompositeSessionIdResolver.java
sdelamo: > You should add `@Indexed(SessionIdResolver.class)` to the interface

done via https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/aafcefb23e3ee3aaceb60a961b7bd22e9bfb8c75

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilter.java
sdelamo: That it is what [`RegexPathMatcher`](https://github.com/micronaut-projects/micronaut-core/blob/4.7.x/core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java) does. It stores it in a concurrent hash map.

security-csrf/src/main/java/io/micronaut/security/csrf/filter/CsrfFilterConfigurationProperties.java
sdelamo: I think the name `regex-pattern` is appropriate. I want to convey is the pattern used for the filter. And I want to convey it is a regular expression not an ant pattern.  

https://github.com/micronaut-projects/micronaut-core/blob/4.7.x/http/src/main/java/io/micronaut/http/annotation/RequestFilter.java#L90

>  The patterns this filter should match

dstepanov (approved):

PR #1847, opened 2024-11-05
fix: Run createAuthentication on a BLOCKING thread

### LINKED ISSUE #1846: NimbusReactiveJsonWebTokenValidator runs on Netty event group threads
### Expected Behavior

`NimbusReactiveJsonWebTokenValidator` should delegate to another thread pool so that `RoleFinder`s etc won't run on a Netty event group thread

### Actual Behaviour

`NimbusReactiveJsonWebTokenValidator` runs on a Netty event group thread.

`JwtTokenProvider` delegated to BLOCKING, but `NimbusReactiveJsonWebTokenValidator` does not. I would expect the behavior not to change. My `RolesFinder` implementation hits a database, so it has blocking operations. At the very least, I should have a way to configure it to use an executor service.

### Steps To Reproduce

Create an application with `io.micronaut.security:micronaut-security-jwt` and io.`micronaut:micronaut-http-server`. Set a breakpoint in the `RolesFinder` resolveRoles function and you will see that it gets called on a Netty thread.

### Environment Information

JDK 21

### Example Application

_No response_

### Version

4.6.6 (core) / 4.6.3 (platform) / micronaut-security-jwt 4.10.2

kevin-wise:
fixes #1846

- matches behavior of NimbusReactiveJsonWebTokenValidator with JwtTokenProvider, which it replaced
- so blocking operations such as a RolesFinder that uses a database won't tie up Netty event loop threads

Let me know if you want this to be configurable, or if it's ok to always do it; and what kind of tests you want.

If it needs to be configurable, would something like this be acceptable?
```java
    NimbusReactiveJsonWebTokenValidator(
            List<GenericJwtClaimsValidator<R>> claimsValidators,
            List<SignatureConfiguration> imperativeSignatureConfigurations,
            List<ReactiveSignatureConfiguration<SignedJWT>> reactiveSignatureConfigurations,
            JsonWebTokenParser<JWT> jsonWebTokenParser,
            ReactiveJsonWebTokenSignatureValidator<SignedJWT> signatureValidator,
            JwtAuthenticationFactory jwtAuthenticationFactory,
            BeanContext beanContext,
            @Value("{" + JwtConfigurationProperties.PREFIX + ".reactive-validator.executor:"
                   + TaskExecutors.BLOCKING + "}") String executor) {
        super(claimsValidators, imperativeSignatureConfigurations, reactiveSignatureConfigurations);
        this.jsonWebTokenParser = jsonWebTokenParser;
        this.signatureValidator = signatureValidator;
        this.jwtAuthenticationFactory = jwtAuthenticationFactory;
        this.scheduler = StringUtils.isEmpty(executor) ? null : getScheduler(beanContext, executor);
    }

    private static Scheduler getScheduler(BeanContext beanContext, String executor) {
        try {
            return Schedulers.fromExecutorService(
                    beanContext.getBean(ExecutorService.class, Qualifiers.byName(executor)));
        } catch (NoSuchBeanException e) {
            throw new ConfigurationException(JwtConfigurationProperties.PREFIX +
                    ".reactive-validator.executor is invalid; no executor configured for name: " +
                    executor, e);
        }
    }

    @Override
    @SingleResult
    public Publisher<Authentication> validateToken(String token, R request) {
        Mono<JWT> publisher = Mono.from(validate(token, request));
        if (scheduler != null) {
            publisher = publisher.subscribeOn(scheduler);
        }
        return publisher
                .map(jwtAuthenticationFactory::createAuthentication)
                .filter(Optional::isPresent)
                .map(Optional::get);
    }
```


kevin-wise:
@sdelamo this is coming from the discussion in https://github.com/micronaut-projects/micronaut-core/discussions/10924

sdelamo:
I think it is correct that `RolesFinder` gets called in the Netty thread as the default implementation `DefaultRolesFinder` is not blocking. 

Honestly, I never thought `RolesFinder` implementation will call a database. 

We could make it configurable to opt-in to `subscribeOn` on the blocking scheduler. But it looks a bit of a hack.

sdelamo:
@kevin-wise I have added a test and a opt-in configuration option. Let me know what you think

kevin-wise:
Thanks for adding the test and the configuration option. I had imagined the configuration would allow you to pick the scheduler, but this is sufficient for my needs.

PR #1848, opened 2024-11-05
fix: Make JwtTokenValidator not a bean

### LINKED ISSUE #1845: Both JwtTokenValidator and NimbusReactiveJsonWebTokenValidator are active and called
### Expected Behavior

When the app is loaded, TokenAuthenticationFetcher should only load NimbusReactiveJsonWebTokenValidator

### Actual Behaviour

When the app is loaded, TokenAuthenticationFetcher loads both NimbusReactiveJsonWebTokenValidator and JwtTokenValidator

### Steps To Reproduce

Create an application with `io.micronaut.security:micronaut-security-jwt` and `io.micronaut:micronaut-http-server`. Set a breakpoint in the constructor of `TokenAuthenticationFetcher` and you will see both beans getting loaded.

https://github.com/micronaut-projects/micronaut-security/pull/1693 seems to be the root cause. JwtTokenValidator is being considered a bean still. Perhaps the `@Inject` on the constructor is causing it to get pulled in?

It might be worth adding a property which would allow each built-in TokenValidator to be disabled. There are certainly use cases where people might want to implement their own and not use the built-in ones.

### Environment Information

JDK 21

### Example Application

_No response_

### Version

4.6.6 (core) / 4.6.3 (platform) / micronaut-security-jwt 4.10.2

kevin-wise:
- remove `@Inject` so JwtTokenValidator will no longer be a bean candidate
- allow disabling NimbusReactiveJsonWebTokenValidator in case of custom implementation

fixes #1845


kevin-wise:
@sdelamo this is coming from the discussion in https://github.com/micronaut-projects/micronaut-core/discussions/10924

sdelamo (approved):
Thanks for the contribution. I have change the configuration properties to be `micronaut.security.token.jwt.nimbus.validator` and `micronaut.security.token.jwt.nimbus.reactive-validator`

kevin-wise:
Looks good. Thanks!

PR #1867, opened 2024-11-20
Context-path prefix added to absolute redirect URLs

ThomasIommi:
Hi again! :wave:

I've noticed an issue in the utility class `ServerContextPathProviderUtils`: the application context path (defined using the property `micronaut.server.context-path`) is always prefixed to all redirect URLs, even when they are configured as absolute URLs. However, when the context path is not set, the URLs remain unaltered.

This behavior results in the following configuration working perfectly:

```yaml
micronaut:
  server:
    port: 8081
  security:
    ... some-oidc-config ...
    redirect:
      enabled: true
      login-success: ${FE_LOGIN_PAGE:`http://localhost:4200/login`}
      login-failure: ${FE_LOGIN_FAILURE_PAGE:`http://localhost:4200/login?oauth2-unauthorized=true`}
      logout: ${FE_LOGOUT_PAGE:`http://localhost:4200/login`}
```
 
But a configuration like this generates invalid redirect URLs, such as _`http://localhost:8081/proxyhttp://localhost:4200/login`_:

```yaml
micronaut:
  server:
    port: 8081
    context-path: /proxy
  security:
    ... some-oidc-config ...
    redirect:
      enabled: true
      login-success: ${FE_LOGIN_PAGE:`http://localhost:4200/login`}
      login-failure: ${FE_LOGIN_FAILURE_PAGE:`http://localhost:4200/login?oauth2-unauthorized=true`}
      logout: ${FE_LOGOUT_PAGE:`http://localhost:4200/login`}
```

Am I misusing the redirect URLs, and would you recommend avoiding absolute URLs?
If not, I have proposed a pull request that attempts to resolve this issue with absolute URLs by checking if the URL includes a scheme. I also tryed to add some unit tests for the changes!

Thank you for your attention and for creating (and maintaining) this amazing framework! :rocket:

### DIFF
--- a/security/src/main/java/io/micronaut/security/config/ServerContextPathProviderUtils.java
+++ b/security/src/main/java/io/micronaut/security/config/ServerContextPathProviderUtils.java
@@ -17,8 +17,11 @@ package io.micronaut.security.config;
 
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.context.ServerContextPathProvider;
 import io.micronaut.http.uri.UriBuilder;
+
+import java.net.URI;
 
 /**
  * Utility methods to prepend a URL with the context path provided via {@link ServerContextPathProvider}.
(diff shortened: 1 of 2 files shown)


ThomasIommi:
___Additional observation___ (not fully related) 

I noticed that the context-path does not seem to be added to the default OAuth2/OIDC endpoints, such as `/oauth/login/keycloak`. To be honest I haven’t looked into the code for this behavior yet, but it might be worth exploring.

If I find this other issue hasn’t already been reported and if you agree, I’ll try to investigate further and, if possible, provide a fix for that as well.

sdelamo (approved):
Thanks for the contribution.

sdelamo:
> f I find this other issue hasn’t already been reported and if you agree, I’ll try to investigate further and, if possible, provide a fix for that as well.

yes, further contributions will be appreciated

PR #1906, opened 2025-01-07
adds session-cookie boolean configuration

### LINKED ISSUE #339: micronaut  cookie max-age   session value
How set `micronaut.security.token.jwt.cookie.cookie-max-age` for session value
eg:
github

sdelamo:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

> Session cookies — cookies without a Max-Age or Expires attribute – are deleted when the current session ends. The browser defines when the "current session" ends, and some browsers use session restoring when restarting. This can cause session cookies to last indefinitely 

Currently, we set an expiration date for the cookie if no expiration date was defined in the configuration. We set the cookie containing the JWT as the value,  with an expiration date matching the JWT `exp` claim. 

Thus, users can't leave the expiration date empty and, therefore, configure a token cookie as a session cookie. [Some users asked for this](https://github.com/micronaut-projects/micronaut-security/issues/339). 

This PR adds a configuration option to define whether a cookie is a session cookie. When set to true, then no Expires attribute is set for the cookie, making it a session cookie.

Close: https://github.com/micronaut-projects/micronaut-security/issues/339


security-jwt/src/test/java/io/micronaut/security/token/jwt/cookie/JwtCookieSessionCookieTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    void testMaxAgeIsSetFromJwtCookieSettings(@Client("/") HttpClient httpClient) {
+        BlockingHttpClient client = httpClient.toBlocking();
+        HttpRequest<?> loginRequest = HttpRequest.POST("/login", Map.of("username","sherlock","password","password"))
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED_TYPE);
+
+        HttpResponse<?> loginRsp = client.exchange(loginRequest, String.class);
+
+        String cookie = loginRsp.getHeaders().get("Set-Cookie");
+        System.out.println(cookie);
```
sdelamo:
```suggestion
```

n0tl3ss (approved):

yawkat (approved):

PR #1916, opened 2025-01-19
Fixed wrong validation annotations

altro3:


altro3:
@sdelamo @graemerocher Review it, please. Found this wrong annotations in micronaut openapi

altro3:
Just so there are no questions: from `@NotBlank` annotation javadoc:
```
The annotated element must not be null and must contain at least one non-whitespace character.
```
In other words, you can't use `@NotBlank` and `@Nullable` together.

altro3:
@sdelamo @graemerocher ping too

sdelamo (approved):
I know that `@NotBlank` means not null and not an empty string. I think we had to add it at one point to ensure compatibility but I don't find any failing test without it. I have removed the `@Nullable` annotation and added a test.

PR #1924, opened 2025-02-03
fix: JWKS Request should not be duplicated headers added by filters

### LINKED ISSUE #1881: Default jwk set fetcher implementation leading to "Request Entity Too Large"
### Expected Behavior

When requesting a jwks file we should not be sending excessive headers to the jwks provider, leading to "Request Entity Too Large".  Irrespective on whether we're using the ReactorCacheJwkSetFetcher or the CacheableJwkSetFetcher

### Actual Behaviour

When upgrading to the latest micronaut version, without defining the jwks cache (basically the default) the ReactorCacheJwkSetFetcher appears to be used.  When using this alongside any filter that modifies the request headers (in our instance the filter is provided by opentelemetry) we end up continually appending extra headers to the request and will eventually cause the server hosting the jwks file to reject our requests.

### Steps To Reproduce

Fresh application, with a dependency on micronaut-security-jwt.

Create one filter to simply log the headers as the final step:
```
@Filter(pattern = "/jwks.json")
@Order(Integer.MAX_VALUE)
public class LogHeadersFilter implements HttpClientFilter {
  private final Logger log = LoggerFactory.getLogger(getClass());
  public Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> rrequest, ClientFilterChain chain) {
    log.info("request: {}, headers= {}", request.getUri(), request.getHeaders().asMap());
    return chain.proceed(request);
  }
}
```

and a simple POC to add a header to the request (in our real apps, this is opentelemetry traceparent)

```
@Filter(pattern = "/jwks.json")
public class AddHeaderFilter implements HttpClientFilter {
  private final Logger log = LoggerFactory.getLogger(getClass());
  public Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> rrequest, ClientFilterChain chain) {
    log.info("Add custom header");
    return chain.proceed(request.header("hdr", "a really long string ..."));
  }
}
```

Running the application with the above, and then hitting an authenticated endpoint, we see in the logs the header "hdr" has multiple copies of the value, and the number of copies increases with each log .  Until eventually the server exposing the jwks file returns an error indicating the "Request Entity Too Large", and our service becomes unusable, requiring a restart.

However, if we set the property:
```
micronaut:
  caches:
    jwks:
      expire-after-write: 60s
```

then we observe the header only has the one value each time it is logged.  The issue here, is that the default behaviour (when blindly upgrading) leads to the exceptions, but requires a new property to be set to avoid this.  So although we have a workaround, we have a fair few services that will require updating, and will likely miss this property on one.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.7.1

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-security/issues/1881

We have two implementations of `JwkSetFetcher`: `CacheableJwkSetFetcher` and `ReactorJwkSetFetcher`. `JwkSetFetcher`defines a bean responsible for sending an HTTP request to a remote authorization server to fetch the public keys, the JSON Web Key Set (JWKS). 

The former is not affected, but the latter, which uses Project Reactor `Mono::cacheInvalidateIf` to cache requests, had a bug that manifested if a client filter added extra HTTP headers to a request sent to the authorization server. For example, if you were using Micronaut Open Telemetry integration. The HTTP Client filter in Micronaut Open Telemetry adds HTTP Headers. Before this fix, the header value would keep getting added.

### DIFF
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ReactorCacheJwkSetFetcher.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ReactorCacheJwkSetFetcher.java
@@ -63,9 +63,13 @@ final class ReactorCacheJwkSetFetcher extends DefaultJwkSetFetcher {
 
     private Mono<JwksCacheEntry> jwksCacheEntry(CacheKey cacheKey) {
         return Mono.from(super.fetch(cacheKey.providerName, cacheKey.url()))
-                .defaultIfEmpty(new JWKSet())
-                .map(jwksSet -> instantiateCacheEntry(cacheKey, jwksSet))
-                .cacheInvalidateIf(JwksCacheEntry::isExpired);
+            .defaultIfEmpty(new JWKSet())
+            .map(jwksSet -> instantiateCacheEntry(cacheKey, jwksSet))
+            .doOnNext(entry -> {
+                if (entry.isExpired()) {
+                    cache.remove(cacheKey);
+                }
+            }).cacheInvalidateIf(JwksCacheEntry::isExpired);
     }
 
     private JwksCacheEntry instantiateCacheEntry(CacheKey cacheKey, JWKSet jwkSet) {
@@ -73,7 +77,7 @@ final class ReactorCacheJwkSetFetcher extends DefaultJwkSetFetcher {
                 ? jwksSignatureConfigurations.get(cacheKey.providerName).getCacheExpiration()
                 : JwksSignatureConfigurationProperties.DEFAULT_CACHE_EXPIRATION));
     }
-    
+
     private record CacheKey(String providerName, String url) {
     }
 
--- a/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ReactorCacheJwkSetFetcher.java
+++ b/security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ReactorCacheJwkSetFetcher.java
@@ -62,14 +62,10 @@ final class ReactorCacheJwkSetFetcher extends DefaultJwkSetFetcher {
     }
 
     private Mono<JwksCacheEntry> jwksCacheEntry(CacheKey cacheKey) {
-        return Mono.from(super.fetch(cacheKey.providerName, cacheKey.url()))
+        return Mono.defer(() -> Mono.from(super.fetch(cacheKey.providerName, cacheKey.url())))
             .defaultIfEmpty(new JWKSet())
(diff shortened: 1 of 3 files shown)


security-jwt/src/main/java/io/micronaut/security/token/jwt/signature/jwks/ReactorCacheJwkSetFetcher.java
```diff
@@ -63,17 +63,21 @@ public Publisher<JWKSet> fetch(@Nullable String providerName, @Nullable String u
 
     private Mono<JwksCacheEntry> jwksCacheEntry(CacheKey cacheKey) {
         return Mono.from(super.fetch(cacheKey.providerName, cacheKey.url()))
-                .defaultIfEmpty(new JWKSet())
-                .map(jwksSet -> instantiateCacheEntry(cacheKey, jwksSet))
-                .cacheInvalidateIf(JwksCacheEntry::isExpired);
+            .defaultIfEmpty(new JWKSet())
+            .map(jwksSet -> instantiateCacheEntry(cacheKey, jwksSet))
+            .doOnNext(entry -> {
```
yawkat: i dont think this works, it will only be called once. you should put it into the cacheInvalidateIf parameter instead

sdelamo: If remove the `doOnNext` and I do 

```java
            .cacheInvalidateIf(entry -> {
                if (entry.isExpired()) {
                    cache.remove(cacheKey);
                }
                return entry.isExpired();
            });
``` 

The test fails 7 headers instead of the expected five.

yawkat: The patch doesn't work either, though. As soon as the first expires, you're going to have multiple JWKSets for the same key, defeating the point of the caching.

This class is deprecated, does it need to be fixed? A proper cache implementation with eviction here would be more complicated than this PR

yawkat: there. i found a simple solution after all

yawkat (approved):

sdelamo:
> there. i found a simple solution after all

Thanks. 

> This class is deprecated, does it need to be fixed? 

Yes. it needs to be fixed.

> A proper cache implementation with eviction here would be more complicated than this PR

@yawkat a proper implementation already exists based on Micronaut Cache.

PR #1970, opened 2025-05-06
Fix LoginController to check configured content types against request content type name without parameters

mattlong-finocomp:
The toString method includes parameters like charset. Using getName instead means only the type and subtype (inc extension) are compared.
Associated with https://github.com/micronaut-projects/micronaut-security/issues/1971

### DIFF
--- a/security/src/main/java/io/micronaut/security/endpoints/LoginController.java
+++ b/security/src/main/java/io/micronaut/security/endpoints/LoginController.java
@@ -177,7 +177,7 @@ public class LoginController<B> {
     @SingleResult
     public Publisher<MutableHttpResponse<?>> login(@Valid @Body UsernamePasswordCredentials usernamePasswordCredentials, HttpRequest<B> request) {
         Optional<MediaType> contentTypeOptional = request.getContentType();
-        if (!(contentTypeOptional.isPresent() && loginControllerConfiguration.getPostContentTypes().contains(contentTypeOptional.get().toString()))) {
+        if (!(contentTypeOptional.isPresent() && loginControllerConfiguration.getPostContentTypes().contains(contentTypeOptional.get().getName()))) {
             return Publishers.just(HttpResponse.notFound());
         }
         return Flux.from(authenticator.authenticate(request, usernamePasswordCredentials))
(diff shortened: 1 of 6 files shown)


graemerocher:
is it possible to add a test?

mattlong-finocomp:
> is it possible to add a test?

I've amended the commit.
Looking at https://github.com/micronaut-projects/micronaut-security/pull/1829/commits/15656152916cb64e2442b651a4b869aa0314ad5e I also updated the LogoutController and OauthController to have the same behaviour.

sascha-frinken:
First thing: Thank you for this PR. 

We just ran into this after goLive today. 
After very difficult live debugging (tcpdump). We ended up addiNg this [1] to our configs.  
Beside case insensitive checking I would really like a log entry in case of sending a 404.


[1]  
```yaml
micronaut:
  security:
    enabled: true
    endpoints:
      login:
        enabled: true
        post-content-types:
          - application/json
          - application/json; charset=UTF-8
          - application/json; charset=utf-8
          - application/x-www-form-urlencoded
          - application/x-www-form-urlencoded; charset=utf-8
          - application/x-www-form-urlencoded; charset=UTF-8
```

sascha-frinken:
The error code 404 is questionable, too.  
Maybe 400 would be better? @graemerocher

sdelamo (approved):

sascha-frinken:
@sdelamo 

No log, really? I will create another PR if you want me to.

sdelamo:
> The error code 404 is questionable, too. Maybe 400 would be better? @graemerocher

@sascha-frinken I have created a PR to make it configurable: https://github.com/micronaut-projects/micronaut-security/pull/1976

sascha-frinken:
> I have created a PR to make it configurable

@sdelamo Appreciate. But I think the most important thing is to add a log message. It was really hard to find the reason the login of an API wasn't working anymore after an update. We had to turn of gzip from proxy to docker container to read the tcpdump on the docker host and start to diff working and failing login requests…

sdelamo:
> > I have created a PR to make it configurable
> 
> @sdelamo Appreciate. But I think the most important thing is to add a log message. It was really hard to find the reason the login of an API wasn't working anymore after an update. We had to turn of gzip from proxy to docker container to read the tcpdump on the docker host and start to diff working and failing login requests…

you are 100% right. I will add [a log statement](https://github.com/micronaut-projects/micronaut-security/pull/1976/commits/38351eeeac14ffad25cd44e0fa3224f000aa791e) and I will have this in mind in future changes.

PR #2017, opened 2025-09-15
feat: protected resource metadata

sdelamo:
- It adds [JSONAssert](https://github.com/skyscreamer/JSONassert) to simplify JSON assertions. 
- Adds [WWW-Authenticate response header]( https://datatracker.ietf.org/doc/html/rfc9110#section-11.6.19) support.
- Adds [protected resource metadata](https://datatracker.ietf.org/doc/html/rfc9728)
- Adds  `AuthenticatorCondition` - a condition that evaluates to true if there is any bean of type ` ReactiveAuthenticationProvider`, `io.micronaut.security.authentication.provider.AuthenticationProvider`, or `io.micronaut.security.authentication.AuthenticationProvider`. This ensures Basic authentication related beans are not loaded unless there is a backing authentication provider. 

These changes are necessary to support [MCP Server Authentication](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#standards-compliance).

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/metadata/DefaultProtectedResourceMetadataProvider.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/metadata/DefaultProtectedResourceMetadataProvider.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.security.oauth2.metadata;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.core.util.StringUtils;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.server.util.HttpHostResolver;
+import io.micronaut.runtime.ApplicationConfiguration;
+import io.micronaut.security.oauth2.configuration.OpenIdClientConfiguration;
+import jakarta.inject.Singleton;
+
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Default implementation of {@link ProtectedResourceMetadataProvider}.
+ */
+@Requires(classes = HttpRequest.class)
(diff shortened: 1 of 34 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/metadata/DefaultProtectedResourceMetadataProvider.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Default implementation of {@link ProtectedResourceMetadataProvider}.
+ */
+@Requires(classes = HttpRequest.class)
+@Experimental
+@Singleton
+public class DefaultProtectedResourceMetadataProvider implements ProtectedResourceMetadataProvider<HttpRequest<?>> {
```
sdelamo, line 40: I have made this class `public` intentionally. I've created a raw implementation, and I'd like to allow downstream replacement via the extends and replace pattern.

security/src/main/java/io/micronaut/security/authentication/BasicAuthWwwAuthenticateChallengeProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    BasicAuthWwwAuthenticateChallengeProvider(ApplicationConfiguration applicationConfiguration) {
+        this.applicationConfiguration = applicationConfiguration;
+    }
+
+    @Override
+    @NonNull
+    public String getWwwAuthenticateChallenge(@Nullable HttpRequest<?> request) {
+        WwwAuthenticateChallenge.Builder builder = WwwAuthenticateChallenge.builder()
+            .authScheme(HttpHeaderValues.AUTHORIZATION_PREFIX_BASIC);
+        applicationConfiguration.getName().ifPresent(name -> builder.param(PARAM_REALM, name));
```
graemerocher, line 54: should this be configurable in any other way apart from the application name?

security/src/main/java/io/micronaut/security/authentication/WwwAuthenticateChallenge.java
```diff
@@ -0,0 +1,123 @@
...
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.CollectionUtils;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.List;
+import java.util.LinkedHashMap;
+
+/**
+ *
```
graemerocher, line 27: add more javadoc about what this is and what is role is

sdelamo: see: https://github.com/micronaut-projects/micronaut-security/pull/2017/commits/8f049796a9887b627e41830a8a8e33da3d4301b1

security/src/main/java/io/micronaut/security/authentication/BasicAuthWwwAuthenticateChallengeProvider.java
sdelamo: [I added param  `charset="UTF-8"`](https://github.com/micronaut-projects/micronaut-security/pull/2017/commits/7f9f434401f4a2c811bfc01d672f1ada905c019d). Users can disable it by setting `micronaut.security.basic-auth.www-authenticate` to `false` and providing their own implementation if they don't like the implementation provided by Micronaut Security.

graemerocher (approved):

PR #2019, opened 2025-09-17
feat: allow define the authorization-server via config

sdelamo:
If you are using a custom domain for your authorization server, a feature common in providers such as Auth0, it is not possible to infer it. This PR allows users to set it explicitly.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfiguration.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfiguration.java
@@ -129,7 +129,7 @@ public interface OauthClientConfiguration extends Toggleable {
     }
 
     /**
-     * @return The {@link AuthorizationServer} used by the OpenID Client.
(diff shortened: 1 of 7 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfiguration.java
```diff
@@ -127,4 +127,13 @@ default SecureEndpoint getTokenEndpoint() throws ConfigurationException {
         return getToken().map(secureEndpointConfiguration -> new DefaultSecureEndpoint(secureEndpointConfiguration, DEFAULT_AUTH_METHOD))
                 .orElseThrow(() -> new ConfigurationException("Oauth client "  + getName() + " requires the token endpoint configuration to be set in configuration"));
     }
+
+    /**
+     * @return The {@link AuthorizationServer} used by the OpenID Client.
```
sdelamo:
```suggestion
     * @return The {@link AuthorizationServer} used by the OAuth Client.
```

graemerocher (approved):

PR #2022, opened 2025-09-18
feat: Token Validation via Auth Server UserInfo endpoint

sdelamo:
see: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

This feature is necessary to support a Micronaut application acting as an MCP Server with OAuth authentication.

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
@@ -25,6 +25,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.client.HttpClient;
+import io.micronaut.http.client.HttpClientConfiguration;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.micronaut.security.oauth2.client.OpenIdProviderMetadata;
 import io.micronaut.security.oauth2.configuration.OpenIdClientConfiguration;
@@ -38,9 +39,12 @@ import java.net.URL;
 @Internal
 final class UserInfoClientFactory {
     private final BeanContext beanContext;
+    private final HttpClientConfiguration httpClientConfiguration;
 
-    UserInfoClientFactory(BeanContext beanContext) {
+    UserInfoClientFactory(BeanContext beanContext,
+                          HttpClientConfiguration httpClientConfiguration) {
         this.beanContext = beanContext;
+        this.httpClientConfiguration = httpClientConfiguration;
     }
 
     @EachBean(OpenIdProviderMetadata.class)
@@ -57,7 +61,7 @@ final class UserInfoClientFactory {
     @Singleton
     UserInfoClientTokenValidator createUserInfoClient(UserInfoClientTokenValidatorConfiguration config) {
         try {
-            HttpClient httpClient = beanContext.createBean(HttpClient.class, new URL(config.baseUrl()));
+            HttpClient httpClient = beanContext.createBean(HttpClient.class, new URL(config.baseUrl()), httpClientConfiguration);
             return new UserInfoClientTokenValidator(config.name(), httpClient, config.path());
         } catch (MalformedURLException e) {
             throw new DisabledBeanException("Malformed URL Exception for UserInfo endpoint " + config.baseUrl() + " for " + config.getName());
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
@@ -36,7 +36,7 @@ import java.net.URL;
 
 @Factory
 @Internal
-class UserInfoClientFactory {
+final class UserInfoClientFactory {
     private final BeanContext beanContext;
 
     UserInfoClientFactory(BeanContext beanContext) {
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.security.oauth2.endpoint.userinfo;
+
+import io.micronaut.context.BeanContext;
+import io.micronaut.context.annotation.EachBean;
+import io.micronaut.context.annotation.Factory;
+import io.micronaut.context.annotation.Prototype;
+import io.micronaut.context.exceptions.DisabledBeanException;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.util.StringUtils;
+import io.micronaut.http.client.HttpClient;
+import io.micronaut.inject.qualifiers.Qualifiers;
+import io.micronaut.security.oauth2.client.OpenIdProviderMetadata;
+import io.micronaut.security.oauth2.configuration.OpenIdClientConfiguration;
+import io.micronaut.security.oauth2.configuration.endpoints.EndpointConfiguration;
+import jakarta.inject.Singleton;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+@Factory
+@Internal
+class UserInfoClientFactory {
+    private final BeanContext beanContext;
+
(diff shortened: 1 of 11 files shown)


security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfigurationProperties.java
```diff
@@ -620,7 +620,8 @@ public static class RegistrationEndpointConfigurationProperties extends DefaultE
          * User info endpoint configuration.
          */
         @ConfigurationProperties("user-info")
-        public static class UserInfoEndpointConfigurationProperties extends DefaultEndpointConfiguration { }
+        public static class UserInfoEndpointConfigurationProperties extends DefaultEndpointConfiguration {
+        }
```
sdelamo:
```suggestion
        public static class UserInfoEndpointConfigurationProperties extends DefaultEndpointConfiguration { }
```

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
```diff
@@ -0,0 +1,94 @@
...
+import io.micronaut.security.oauth2.configuration.OpenIdClientConfiguration;
+import io.micronaut.security.oauth2.configuration.endpoints.EndpointConfiguration;
+import jakarta.inject.Singleton;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+@Factory
+@Internal
+class UserInfoClientFactory {
```
graemerocher, line 39: probably can be `final`

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
```diff
@@ -0,0 +1,94 @@
...
+            userInfoEndpoint = metadata.getUserinfoEndpoint();
+        }
+        return createUserInfoClientWithUrl(userInfoEndpoint, metadata.getName());
+    }
+
+    @EachBean(UserInfoClientTokenValidatorConfiguration.class)
+    @Singleton
+    UserInfoClientTokenValidator createUserInfoClient(UserInfoClientTokenValidatorConfiguration config) {
+        try {
+            HttpClient httpClient = beanContext.createBean(HttpClient.class, new URL(config.baseUrl()));
```
graemerocher, line 60: where is the client closed / shutdown?

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
```diff
@@ -0,0 +1,94 @@
...
+    @NonNull
+    private UserInfoClientTokenValidatorConfiguration createUserInfoClientWithUrl(@Nullable String userInfoEndpoint,
+                                                                                  @NonNull String name) {
+        if (StringUtils.isEmpty(userInfoEndpoint)) {
+            throw new DisabledBeanException("UserInfo endpoint not set for " + name);
+        }
+        try {
+            URL url = new URL(userInfoEndpoint);
+            String path = url.getPath();
+            String baseUrl = url.toString().substring(0, url.toString().indexOf(path));
```
graemerocher, line 84: is this URL from user input? If so is it sanitised and prevented from being relative?

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
```diff
@@ -0,0 +1,86 @@
...
+import reactor.core.publisher.Mono;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * {@link TokenValidator} which uses a remote `UserInfo` endpoint to validate a token.
+ */
+@Internal
+class UserInfoClientTokenValidator implements Closeable, TokenValidator<HttpRequest<?>>, Named {
```
graemerocher, line 40: can be `final`

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
graemerocher, line 60: I don't think a new client is needed for each validator. Can you not just use the default client and pass absolute URLs to `retrieve`?

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
```diff
@@ -0,0 +1,86 @@
...
+            .flatMap(m -> {
+                Object subject = m.get(Claims.SUBJECT);
+                if (subject == null) {
+                    return Mono.empty();
+                }
+                return Mono.just(Authentication.build(subject.toString(), m));
+            })
+            .onErrorResume(t -> {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(t.getMessage(), t);
```
graemerocher, line 76: add more detail to this error message of where this went wrong

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void close() throws IOException {
+        httpClient.close();
+    }
+
+    @Override
+    @NonNull
+    public Publisher<Authentication> validateToken(@NonNull String token, @Nullable HttpRequest<?> request) {
+        return Mono.from(httpClient.retrieve(HttpRequest.GET(path).bearerAuth(token),
+                Argument.mapOf(String.class, Object.class)))
```
graemerocher, line 66: extract this `mapOf` to a constant

graemerocher (changes requested):

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/2022/commits/0ee1c7bd985eea1a6b2e543b2e54fbc01d729edf

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/2022/commits/90d76f0c64d1c05a3b89b43ff060dd935d5d78eb

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/2022/commits/f2fc4ac034c6f512a099368630907dd5a04027cc

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
sdelamo: the client is shutdown in `UserInfoClientTokenValidator::close`

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
sdelamo: Typically, the `UserInfo` endpoint is fetched from `.well-known/openid-configuration`. For example, for Google, it is exposed here: https://accounts.google.com/.well-known/openid-configuration, and it has a value of https://openidconnect.googleapis.com/v1/userinfo.

However, if, for whatever reason, the auth server does not populate that `UseInfo` endpoint correctly, or the user needs to set it manually, we allow them to set it via configuration. For example, the user could set `micronaut.security.oauth2.clients.authserver.openid.user-info.url` and the configuration value takes precedence over the `.well-known/openid-configuration` value.

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientTokenValidator.java
sdelamo: An error is expected. I set the level to `debug` to avoid filling the application logs. For example, for an invalid token, we expect to get a 401 response for the `/userinfo` endpoint, and Micronaut HTTP Client will wrap it in an `HTTPClientResponseException`.

sdelamo: I added some extra logs in `debug` and move the generic logs to `trace`

https://github.com/micronaut-projects/micronaut-security/pull/2022/commits/fb998447b653b1c256b444d65a62836f3c247e82

security-oauth2/src/main/java/io/micronaut/security/oauth2/endpoint/userinfo/UserInfoClientFactory.java
sdelamo: What do you mean by the `default client`? There will typically be only one `UserInfoClientTokenValidator`   HTTP Client, as most applications will only have one OAuth2 client configured.

sdelamo: I passed the HTTP Client configuration so that users are able to configure it https://github.com/micronaut-projects/micronaut-security/pull/2022/commits/9e2e1027e82b78e8013abab34566c03f37e245e7

graemerocher (approved):

PR #2089, opened 2025-12-17
Security OAuth2: Support Keycloak >= 18 logout params via configurable provider version

knobo:
This PR improves Keycloak RP-initiated logout handling in `KeycloakEndSessionEndpoint` by switching between legacy and modern logout parameters based on a configurable provider version.

#### What changed

- Added `OauthClientConfiguration#getVersion()` (default `null`) to represent the **provider/server version** used for feature support decisions.
    
- Implemented `version` property in `OauthClientConfigurationProperties` with getter/setter for configuration binding.
    
- Updated `KeycloakEndSessionEndpoint`:
    
    - Uses `redirect_uri` for Keycloak versions **< 18.0.0**.
        
    - Uses `post_logout_redirect_uri` for Keycloak versions **\>= 18.0.0**.
        
    - Adds `id_token_hint` if available, otherwise falls back to `client_id` and logs a debug message.
        
    - Uses Micronaut `SemanticVersion` parsing with a safe default (invalid/blank version → modern behavior).
        

#### Why

Keycloak 18 introduced/standardized `post_logout_redirect_uri` for RP-initiated logout flows, while older versions use `redirect_uri`. This change enables correct parameter usage without hardcoding assumptions in the endpoint.

#### Configuration

Example:

```yaml
micronaut:
  security:
    oauth2:
      clients:
        keycloak:
          version: "18.0.0"
```

#### Backward compatibility question

This change preserves backward compatibility with Keycloak versions prior to 18 by switching behavior based on the configured provider version.

That said, I’d like feedback on whether supporting legacy logout behavior is still desirable at all. Keycloak is a security-critical component, and older versions that require `redirect_uri` are no longer security-maintained upstream.

If Micronaut Security prefers to:

- **require modern Keycloak versions**, or
    
- **drop legacy logout support entirely**, or
    
- **support both only for a limited transition period**
    

I’m happy to adjust the implementation accordingly (including simplifying the logic and/or documentation).

### DIFF
--- a/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfiguration.java
+++ b/security-oauth2/src/main/java/io/micronaut/security/oauth2/configuration/OauthClientConfiguration.java
@@ -147,4 +147,13 @@ public interface OauthClientConfiguration extends Toggleable {
     default boolean isProxyWellKnownOpenidConfiguration() {
         return false;
     }
+
+    /**
+     * @return The version of the provider (e.g. "18.0.0"). Used to determine feature support.
+     * @since 5.0.0
+     */
+    @Nullable
+    default String getVersion() {
+        return null;
+    }
 }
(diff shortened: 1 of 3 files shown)


knobo:
Now, I see that there is already done some work on that here #1049
But I think it is important to move this issue forward, as Keycloak should be up-to-date and security patched, and actually I think we should drop backward compatibility.

knobo:
Solved by this: #2090

PR #2116, opened 2026-01-19
Add AuthenticationWebSocketStateBinder for WebSockets

zManu3k:
Introduces AuthenticationWebSocketStateBinder to resolve Authentication from the WebSocket state instead of requiring manual userPrincipal access.

### DIFF
--- a/security-websocket/build.gradle.kts
+++ b/security-websocket/build.gradle.kts
@@ -1,0 +1,13 @@
+plugins {
+    id("io.micronaut.build.internal.security-module")
+}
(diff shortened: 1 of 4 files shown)


graemerocher:
needs tests

sdelamo:
Close via: https://github.com/micronaut-projects/micronaut-security/pull/2128

Thanks for the contribution @zManu3k 

I moved your change to #2128 instead of creating a new module. I added it with `@Requires` and `compileOnly`.

PR #2125, opened 2026-02-05
Add Base64 encoding support for SecretEncryptionConfiguration

MariusVolkhart:
Mirrors the base64 encoding functionality that exists in `SecretSignatureConfiguration` to `SecretEncryptionConfiguration`, allowing JWT encryption secrets to be supplied as Base64-encoded strings.

Changes:
- Add base64 field with getter/setter to `SecretEncryptionConfiguration`
- Modify `SecretEncryption` constructor to decode Base64 secrets when the base64 flag is set (default: false)
- Add comprehensive integration tests (`JwtEncryptionBase64Spec`) covering Base64 encoding, decryption, and backward compatibility
- Add documentation (jwtEncryptionSecret.adoc) with examples of both plain text and Base64-encoded secret configuration
- Update `SecretEncryptionSpec` with YAML configuration example

All examples use A256GCM encryption method with dir (direct) algorithm for modern AEAD-based encryption.

Backward compatibility is fully maintained with base64 defaulting to false.


sdelamo (approved):
Thanks for the contribution @MariusVolkhart

PR #2209, opened 2026-05-26
allow resolving token roles from nested Authentication attributes

### LINKED ISSUE #1132: Get role from nested value in Authentication attributes map
### Feature description

The property `micronaut.security.token.roles-name` allows you to specify  the authentication attributes map key for the user’s roles.

But my JWT token payload looks like this:
```json
...
"realm_access": {
    "roles": [
      "manager",
      "offline_access",
      "uma_authorization"
    ]
  },
...
```

It would be useful to be able to specify a nested key. In this case that would be:
`micronaut.security.token.roles-name: realm_access.roles`

Relevant code can be found here: https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/token/DefaultRolesFinder.java#L81

A simular question (with workaround) has been posted on the micronaut gitter channel: https://gitter.im/micronautfw/questions?at=5e4501aab612cc7bb1643207

kkriske:
Allow resolving roles from nested authentication attributes, by setting the `micronaut.security.token.roles-name-separator` property.

This also fixes a potential bug where the `micronaut.security.token.roles-separator` property was interpreted as a regex, through `String#split`, where I think that is unintended behaviour and it should be seen as a literal separator.

I have annotated a few things I wish to point out and/or desire feedback on.

fixes #1132


security-jwt/src/main/java/io/micronaut/security/token/jwt/generator/claims/JWTClaimsSetGenerator.java
```diff
@@ -182,7 +183,17 @@ protected void populateWithAuthentication(JWTClaimsSet.Builder builder, Authenti
         if (!rolesKey.equalsIgnoreCase(TokenConfiguration.DEFAULT_ROLES_NAME)) {
             builder.claim(ROLES_KEY, rolesKey);
```
kkriske, line 184: The `ROLES_KEY` claim still ignores the separator and I'm not sure what to do about it. I can add a `ROLES_KEY_SEPARATOR` claim but that feels the incorrect approach.
This is a non-standard claim (I think) that's only used by the `ClientAuthentication` class, I see no other documented use-cases.

Should I make this a list property in case a separator is set, and handle it in the ClientAuthentication like that?

security/src/main/java/io/micronaut/security/token/DefaultRolesFinder.java
```diff
@@ -56,28 +56,40 @@ public DefaultRolesFinder(TokenConfiguration tokenConfiguration) {
      */
     @NonNull
     private List<String> rolesAtObject(@Nullable Object rolesObject) {
-        if (rolesObject == null) {
-            return emptyList();
-        }
+        return switch (rolesObject) {
+            case null -> emptyList();
+            case CharSequence _ when tokenConfiguration.getRolesSeparator() != null ->
+                asList(rolesObject.toString().split(Pattern.quote(tokenConfiguration.getRolesSeparator())));
```
kkriske, line 62: I think this was a pre-existing bug, where the separator was passed straight to `String#split`. It seems like an unintended side-effect that this means the separator is a regex rather than a string literal.

Please advise on what the intention would be.

security/src/main/java/io/micronaut/security/token/DefaultRolesFinder.java
```diff
@@ -56,28 +56,40 @@ public DefaultRolesFinder(TokenConfiguration tokenConfiguration) {
      */
     @NonNull
     private List<String> rolesAtObject(@Nullable Object rolesObject) {
-        if (rolesObject == null) {
-            return emptyList();
-        }
+        return switch (rolesObject) {
```
kkriske, line 59: Except for the `Pattern#quote` call, this is just a rewrite to use fancy new syntax. Can revert if unwanted.
Use of switch expressions also makes back-porting harder so I can avoid using jdk17+ features for that reason as well if desired.

security/src/main/java/io/micronaut/security/token/config/TokenConfigurationProperties.java
```diff
@@ -45,6 +45,9 @@ public class TokenConfigurationProperties implements TokenConfiguration {
     @NonNull
     private String nameKey = TokenConfiguration.DEFAULT_NAME_KEY;
 
+    @Nullable
+    private String rolesNameSeparator = TokenConfiguration.DEFAULT_ROLES_NAME_SEPARATOR;
```
kkriske, line 49: I do think the additional property is a bit confusing, but I think this is the only way of introducing this functionality without breaking backwards compatibility.

E.g. is it possible to change the `rolesName` property to a list without introducing subtle changes? e.g. a pre-existing value of `a,b,c` would then automatically split into a List I think? Is there a way to disable parsing config values into lists, so a `property: value` always parses into a singleton list?
Not to mention the actual breaking change of changing the type of the field, so that's a no-go either way.

kkriske:
@graemerocher would you mind reviewing the things I pointed out? And whether this is generally in the right direction to be accepted?

sdelamo (commented):
I don't think we should do this change. 

The settings are not ease to comprehend after this change. There will be two quite confusing names `micronaut.security.token.roles-name-separator`,  `micronaut.security.token.roles-separator`. 

There is already an API to ease custom Role parsing logic. You can create you own replacement for the [`RolesFinder` AP](https://micronaut-projects.github.io/micronaut-security/5.0.0/api/io/micronaut/security/token/RolesFinder.html). 

```java
@Replaces(RoleFinder.class)
@Singleton
class RoleFinderReplaces implements RoleFinder {
// Implement the API
}
```

I will create a separate documentation PR documenting better the [`RolesFinder` AP](https://micronaut-projects.github.io/micronaut-security/5.0.0/api/io/micronaut/security/token/RolesFinder.html).

kkriske:
@sdelamo If there's a way to allow a breaking change, perhaps for 5.1.x, to make `micronaut.security.token.roles-name` a `List<String>` instead of a `String`, the configuration complexity would stay on par.

While replacing the RoleFinder works fine, I think fairly standard use-cases should be covered by default.
Right now, Micronaut Test Resources supports Keycloak out of the box https://micronaut-projects.github.io/micronaut-test-resources/latest/guide/#modules-oauth2, but using it requires this custom RolesFinder implementation.

sdelamo:
see: https://github.com/micronaut-projects/micronaut-security/pull/2218 

I will check Keycloak integration. Thanks for this PR nonetheless and starting the conversation.

PR #2220, opened 2026-06-11
Support @RunAs authentication delegation

sdelamo:
This pull-request adds support for a new annotation: `RunAs ` and the Jakarta `jakarta.annotation.security.RunAs` annotation by mapping the annotation value to an Authentication and applying it to the security context for the scope of the intercepted method invocation.


sdelamo:
I have some doubt about whether we should use the Jakarta annotation `jakarta.annotation.security.RunAs` or create a new Micronaut Security specific annotation.

`jakarta.annotation.security.RunAs` targets only TYPE, and the annotation javadoc says:

> Defines the identity of the application during execution. This allows developers to execute an application under a particular role. The role must map to the user / group information in the container’s security realm. Its value is the name of a security role.

It has been used in the past to specify only a role instead of an entire authentication, as I did in this PR.

graemerocher:
yeah not sure this makes sense, `@RunAs` the value has to be a role

security/src/main/java/io/micronaut/security/authentication/RunAsInterceptor.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        ServerRequestContextSecurityContextSupplier.ScopedSecurityContext scopedContext = ServerRequestContextSecurityContextSupplier.openSecurityContext();
+        SecurityContext securityContext = scopedContext.getSecurityContext();
+        Authentication previousAuthentication = securityContext.getAuthentication();
+        String previousToken = securityContext.getToken();
+        Integer previousRejectionStatus = securityContext.getRejectionStatus();
+        try {
+            securityContext.withAuthentication(runAsAuthentication);
+            return context.proceed();
```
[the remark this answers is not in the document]
sdelamo: I think I fixed here: https://github.com/micronaut-projects/micronaut-security/pull/2220/commits/aecab4b150bd7f55b10ab9d0adb0b510a5be8ae9

sdelamo (commented):
@dstepanov, I am interested to know whether you think the changes to security/src/main/java/io/micronaut/security/context/ServerRequestContextSecurityContextSupplier.java are sound. We need to support use cases where there is no request.

For example: security/src/test/java/io/micronaut/security/authentication/RunAsScheduledTest.java

security-annotations/src/main/java/io/micronaut/security/annotation/Attribute.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * An attribute entry for {@link RunAs}.
+ *
+ * @author Sergio del Amo
+ * @since 5.1.0
+ */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Attribute {
```
graemerocher: since this annotations name is quite generic, should it be nested inside RunAs?

graemerocher (approved):

security-annotations/src/main/java/io/micronaut/security/annotation/Attribute.java
sdelamo: Yes, good point. Since it is only valid as a RunAs member and Attribute is too broad as a top-level annotation name, I’ll move it to RunAs.Attribute and update the examples/tests.

Done via https://github.com/micronaut-projects/micronaut-security/pull/2220/commits/adc3ea633a2588eabd2fe5e9b3a2a0477acb57a5

PR #2238, opened 2026-07-02
Resolve locale from OIDC profile claims

sdelamo:
Add `ProfileClaims` constants for the [OpenID Connect standard profile claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) and use the locale claim to [resolve the HTTP locale](https://docs.micronaut.io/5.0.4/guide/index.html#localeResolution) from the current authentication.

Introduce a `LocaleClaimHttpLocaleResolver` ordered after cookie/session locale resolution and before request header resolution. Cover missing, invalid, string, and Locale claim values with unit tests, and add a JWT functional test that verifies authenticated requests can localize responses from the token locale claim.

### DIFF
--- a/security/src/main/java/io/micronaut/security/token/LocaleClaimHttpLocaleResolver.java
+++ b/security/src/main/java/io/micronaut/security/token/LocaleClaimHttpLocaleResolver.java
@@ -35,7 +35,7 @@ import java.util.Optional;
  * current {@link Authentication}.
  *
  * @author Sergio del Amo
- * @since 5.2.1
+ * @since 5.3.0
  */
 @Requires(classes = HttpRequest.class)
 @Requires(beans = HttpLocaleResolutionConfiguration.class)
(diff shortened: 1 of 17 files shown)


security/src/main/java/io/micronaut/security/token/LocaleClaimHttpLocaleResolver.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Resolves the HTTP locale from the OpenID Connect {@code locale} claim in the
+ * current {@link Authentication}.
+ *
+ * @author Sergio del Amo
+ * @since 5.2.1
```
radovanradic: The class below has `5.3.0` so this probably should be the same (if important).

radovanradic (approved):

security/src/main/java/io/micronaut/security/token/LocaleClaimHttpLocaleResolver.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/2238/commits/2bd8c468a13455e95a860762e470548450d4f2c8

PR #2245, opened 2026-07-08
 Improve OAuth token URL configuration errors

### LINKED ISSUE #2179: Improve Error messages on OAuth verification
### Feature description

On a production service we had an error with this message `io.micronaut.discovery.exceptions.NoAvailableServiceException: No available services for ID: Load balancer contains no servers` and it ended up been related to a missing `https://` specifier on the `micronaut.security.oauth2.clients.*.token.url`

The error was really hard to trace to that point. I would love to have improved errors for those cases

lissam1:
Fixes #2179

  Adds validation for configured OAuth token endpoint URLs so invalid values fail with a clear ConfigurationException.
  Changes:

  - Validate micronaut.security.oauth2.clients.*.token.url when building the token endpoint.
  - Require an absolute http or https URL with a valid host.
  - Include the OAuth client name, property path, and configured value in the error message.
  - Add regression coverage for missing schemes, malformed HTTP URLs, relative paths, and unsupported schemes.


sdelamo (changes requested):
we should do this with an annotation constraint.

PR #2258, opened 2026-07-22
Add RFC 7643 SCIM core schema support

sdelamo:
[RFC 7643](https://datatracker.ietf.org/doc/html/rfc7643)

Introduce the security-scim-core module with Java models for SCIM User, Group, enterprise User extensions, and discovery resources.

Add Micronaut Serialization and Jakarta Validation support, including schema URI handling and custom extensions. Cover every serializable model with introspection, validation, and RFC-derived JSON deserialization tests.

Document installation, resource models, schema extensions, discovery resources, and the boundary between RFC 7643 models and RFC 7644 protocol operations.

Every class is currently annotated with @Experimental.


sdelamo:
@graemerocher there is an [open source Java library for SCIM](https://github.com/pingidentity/scim2/tree/master) which we could use instead. However, it is coupled to [Jackson Databind](https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/pom.xml#L231-L234)

PR #2266, opened 2026-08-19
Content Security Policy (CSP) module

sdelamo:
This PR adds a more robust CSP support. I will deprecates the [CSP support  of Micronaut Views](https://micronaut-projects.github.io/micronaut-views/latest/guide/index.html#csp)


security-csp/src/main/java/io/micronaut/security/csp/filters/ContentSecurityPolicyFilter.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Generates and stores the nonce before view rendering and response policy generation when
+     * nonce support is enabled for {@code script-src}.
+     *
+     * @param request the current request
+     */
+    @RequestFilter
+    void generateNonce(HttpRequest<?> request) {
+        if (scriptSrcConfiguration.isEnabled() && scriptSrcConfiguration.isNonce()) {
```
dstepanov, line 73: A nonce is only generated when `script-src` has nonce enabled, but `StyleSrcConfiguration` also implements `NonceConfiguration` and `DefaultContentSecurityPolicyGenerator.styleSrc(request)` looks for the same request attribute.

With `micronaut.security.csp.style-src.nonce=true` and `script-src.nonce` left at its default `false`, no nonce is ever stored on the request. `directive()` then finds no `cspNonce` attribute, the value list ends up empty, and because `none` defaults to `true` the response gets `style-src 'none'` — every stylesheet/inline style is blocked and the view has no nonce to use.

Suggest checking every nonce-capable directive configuration:

```suggestion
        if ((scriptSrcConfiguration.isEnabled() && scriptSrcConfiguration.isNonce())
            || (styleSrcConfiguration.isEnabled() && styleSrcConfiguration.isNonce())) {
```
(injecting `StyleSrcConfiguration` alongside `ScriptSrcConfiguration`).

---

security-csp/src/main/java/io/micronaut/security/csp/filters/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * HTTP server filter support for Content Security Policy response headers.
+ */
+@NullMarked
+@Requires(classes = ServerFilter.class)
+@Requires(property = ContentSecurityPolicyFilterConfigurationProperties.PROPERTY_ENABLED, value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
```
dstepanov, line 21: This package is gated only on `micronaut.security.csp.filter.enabled`, not on `micronaut.security.csp.enabled`.

`io.micronaut.security.csp/package-info.java` gates the whole generator package on `micronaut.security.csp.enabled`. So with `micronaut.security.csp.enabled=false`, `DefaultContentSecurityPolicyGenerator` disappears but `ContentSecurityPolicyFilter` is still a bean, and its constructor requires a `ContentSecurityPolicyGenerator`. Filter beans are resolved when the server builds its filter chain, so the application fails with `NoSuchBeanException` rather than just running without CSP. There is no test for `csp.enabled=false`.

Adding `@Requires(property = ContentSecurityPolicyConfigurationProperties.PREFIX + ".enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)` here (or `@Requires(beans = ContentSecurityPolicyGenerator.class)` on the filter) would fix it.

---

security-csp/src/main/java/io/micronaut/security/csp/DefaultContentSecurityPolicyGenerator.java
```diff
@@ -0,0 +1,609 @@
...
+        directives.add(reportUri());
+        directives.add(requireTrustedTypesFor());
+        directives.add(frameAncestors());
+        directives.add(frameSrc());
+        directives.add(imgSrc());
+        directives.add(manifestSrc());
+        directives.add(mediaSrc());
+        directives.add(formAction());
+        directives.add(workerSrc());
+        return directives.stream().filter(Objects::nonNull).toList();
```
dstepanov, line 291: Five directive configurations shipped in this PR are never read by the generator: `ChildSrcConfiguration`, `ScriptSrcAttrConfiguration`, `ScriptSrcElemConfiguration`, `StyleSrcAttrConfiguration` and `StyleSrcElemConfiguration`.

They are full `@ConfigurationProperties` beans (`micronaut.security.csp.child-src.*`, `.script-src-attr.*`, `.script-src-elem.*`, `.style-src-attr.*`, `.style-src-elem.*`), and `ContentSecurityPolicy` even exposes `childSrc()`, `scriptSrcElem()` etc. for reading them back — but neither `directives()` nor `directives(HttpRequest)` emits them, so those directives never appear in the header.

Concretely: a user who sets `micronaut.security.csp.script-src-elem.self=true` gets no `script-src-elem` directive at all, and their scripts stay blocked by the default `script-src 'none'`. Note `style-src-elem` also has a `nonce` flag that can never take effect.

---

security-csp/src/test/java/io/micronaut/security/csp/views/CspViewModelProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+ * Adds the request's CSP nonce to map-backed view models.
+ *
+ * <p>When a nonce is available, it is exposed as {@code cspNonce} so a view can apply it to
+ * trusted {@code script} elements. The processor copies the model before adding the value, so it
+ * also supports immutable input maps.</p>
+ */
+@Requires(classes = ViewModelProcessor.class)
+@Singleton
+@Internal
+final class CspViewModelProcessor implements ViewModelProcessor<Map<String, Object>, HttpRequest<?>> {
```
dstepanov, line 39: This class lives in `src/test`, so it is not part of the published module — but the documentation promises it as a module feature.

`contentSecurityPolicy.adoc`: "When nonce support and Micronaut Views are present, it also makes the request nonce available to map-backed view models", and `cspNonceGeneration.adoc`: "the module also adds the nonce to the model with that key". `build.gradle.kts` even declares `compileOnly(mnViews.micronaut.views.core)` for main sources, but no main-source class implements `ViewModelProcessor`.

Result: an application that enables `script-src.nonce=true` and renders `${cspNonce}` gets an unresolved model key, so the inline script has no nonce and is blocked by the policy. Only this module's own tests work, because they define the processor themselves. This should move to `src/main/java`.

---

security-csp/src/main/java/io/micronaut/security/csp/report/ContentSecurityPolicyHashReportBody.java
```diff
@@ -0,0 +1,41 @@
...
+ * @param hash cryptographic digest prefixed by its algorithm
+ * @param destination request destination, such as {@code script}
+ * @param type kind of resource that was hashed, currently {@code subresource}
+ * @see <a href="https://w3c.github.io/webappsec-csp/#csp-hash-report">CSP hash reports</a>
+ * @since 5.4.0
+ */
+@Serdeable
+public record ContentSecurityPolicyHashReportBody(
+    @NotBlank @JsonProperty("document_url") String documentUrl,
+    @NotBlank @JsonProperty("subresource_url") String subresourceUrl,
```
dstepanov, line 36: These snake_case keys look wrong. `CSPHashReportBody` is a WebIDL interface whose `toJSON()` serializes its members verbatim — `documentURL`, `subresourceURL`, `hash`, `type`, `destination` — so a browser sends camelCase, not `document_url`/`subresource_url`. (Only the Reporting API *envelope* uses snake_case, hence `user_agent` in `SerializedContentSecurityPolicyReport`, which is correct.)

Note the sibling `ContentSecurityPolicyViolationReportBody` in the same package uses `@JsonProperty("documentURL")` for the same concept.

With the current mapping a real `csp-hash` delivery leaves both fields null, the `@NotBlank` constraints fail and the controller returns 400 instead of 204, so the handler never sees the report. `ContentSecurityPolicyControllerTest` doesn't catch this because its fixture was written against the implementation.

```suggestion
    @NotBlank @JsonProperty("documentURL") String documentUrl,
    @NotBlank @JsonProperty("subresourceURL") String subresourceUrl,
```

---

security-csp/src/main/java/io/micronaut/security/csp/DefaultContentSecurityPolicyGenerator.java
```diff
@@ -0,0 +1,609 @@
...
+    }
+
+    /**
+     * Builds the configured deprecated {@code report-uri} directive.
+     *
+     * @return the configured deprecated {@code report-uri} directive, or {@code null} when disabled
+     * @since 5.4.0
+     */
+    protected @Nullable ContentSecurityPolicyDirective reportUri() {
+        return cspConfiguration.isReportUriEnabled() ? directive(REPORT_URI, cspConfiguration.getReportUri()) : null;
```
dstepanov, line 399: No guard for an empty endpoint list. `getReportUri()` defaults to `Collections.emptyList()`, so `micronaut.security.csp.report-uri-enabled=true` alone produces `new ContentSecurityPolicyDirective("report-uri", "")`, whose `toString()` emits the bare token `report-uri` — a value-less directive that browsers report as a CSP parse error.

The `DirectiveConfiguration` overload handles the equivalent case by returning `null`; this path should too:

```suggestion
        return cspConfiguration.isReportUriEnabled() && CollectionUtils.isNotEmpty(cspConfiguration.getReportUri())
            ? directive(REPORT_URI, cspConfiguration.getReportUri()) : null;
```

---

dstepanov (commented):
Correctness review of the new CSP module. Six findings, two of them likely to break applications: `style-src` nonces are never generated, and `micronaut.security.csp.enabled=false` leaves the filter with an unsatisfiable dependency. Also flagged: five directive configurations that are bound but never emitted, the view-model nonce processor living in test sources while the docs promise it, snake_case JSON keys on the `csp-hash` report body, and a value-less `report-uri` when the endpoint list is empty.

---

PR #2271, opened 2026-08-21
Add fetch metadata module

sdelamo:
https://web.dev/articles/fetch-metadata#enforcing_a_resource_isolation_policy


security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/CrossOriginFetchMetadataRule.java
```diff
@@ -0,0 +1,94 @@
...
+        return !allowedOrigins.isEmpty() &&
+            ((configuration.getAllowedOriginsRegex().isEmpty() &&
+                Objects.equals(allowedOrigins, CorsOriginConfiguration.ANY)) ||
+                allowedOrigins.contains(origin));
+    }
+
+    static boolean matchesMethod(CorsOriginConfiguration configuration, HttpMethod method) {
+        List<HttpMethod> allowedMethods = configuration.getAllowedMethods();
+        return Objects.equals(allowedMethods, CorsOriginConfiguration.ANY_METHOD) ||
+            allowedMethods.contains(method);
```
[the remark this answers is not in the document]
sdelamo: this should be addressed via https://github.com/micronaut-projects/micronaut-security/pull/2271/commits/6174fbd0bcf5785b86d8ea5f7544ed1a9cc770b4

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/FetchMetadataRulesConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+     * Enabled by default.
+     *
+     * @return whether same-origin requests are allowed
+     */
+    boolean isAllowSameOrigin();
+
+    /**
+     * Whether requests from another origin on the same site are allowed.
+     * Applications should enable this only when every same-site origin is trusted.
+     * Disabled by default.
```
yawkat: Add a note that even when disabled, such requests are allowed if there is matching cross-origin config.

sdelamo: I changed the javadoc https://github.com/micronaut-projects/micronaut-security/pull/2271/commits/2d1abb2b9b23d9ae7afdce14473952ee84ee8946

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/NoFetchMetadataRule.java
```diff
@@ -0,0 +1,51 @@
...
+@Internal
+final class NoFetchMetadataRule implements HttpRequestFetchMetadataRule {
+    @Override
+    public int getOrder() {
+        return FetchMetadataRuleOrder.NO_FETCH_METADATA;
+    }
+
+    @Override
+    public FetchMetadataRuleResult check(HttpRequest<?> request, @Nullable SecFetch secFetch) {
+        if (secFetch == null) {
```
dstepanov: **Fail-open on unrecognized values, not just absent ones.**

`SecFetch.of()` returns `null` if **any** of `Sec-Fetch-Site` / `-Mode` / `-Dest` carries a value not in the corresponding enum (and also when `Sec-Fetch-User` is present with a value other than `?1`). Since this rule runs first (order 100) and is enabled by default, a real browser request carrying a destination micronaut-http doesn't know is fully exempted from the policy instead of being evaluated.

That isn't hypothetical: `io.micronaut.http.Destination` currently has no `webidentity` (FedCM), `dictionary`, `webbundle`, or `sharedstorageworklet`, and the Fetch spec's destination list only grows. Every future browser addition silently widens the hole, and nothing here will fail loudly when it happens.

`Sec-Fetch-Site` alone is the signal for "client is not Fetch-Metadata-aware". Suggest keying on its absence directly:

```java
if (!request.getHeaders().contains(HttpHeaders.SEC_FETCH_SITE)) {
    return FetchMetadataRuleResult.ALLOWED;
}
return FetchMetadataRuleResult.UNKNOWN;
```

A request with `Sec-Fetch-Site: cross-site` and an unparseable `Dest` then falls through to the remaining rules and gets rejected — fail-closed. The docs do describe today's behaviour ("absent, incomplete, and unrecognized"), so this is a deliberate call rather than an oversight, but I think it lands on the wrong side of the trade-off for a security default.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/SimpleTopLevelNavigationAndIframingFetchMetadataRule.java
```diff
@@ -0,0 +1,54 @@
...
+import io.micronaut.security.fetchmetadata.HttpRequestFetchMetadataRule;
+import jakarta.inject.Singleton;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Allows simple {@code GET} navigations, including document and frame navigations, while
+ * excluding requests whose destination is {@code object} or {@code embed}.
+ */
+@Internal
+@Singleton
```
dstepanov, line 40: **No enable/disable toggle, unlike the other five built-in rules.**

Every sibling rule has a `@Requires(property = ...)` plus a matching accessor on `FetchMetadataRulesConfiguration`. This one has neither — and it's also missing the `@Requires(classes = HttpRequest.class)` guard the others carry.

That matters because this is the most permissive built-in rule: it allows cross-site `GET` navigations including `frame`/`iframe` destinations. It's exactly the rule an application hardening beyond the default policy would want to switch off, and right now there's no way to do that short of not using the module.

Suggest `micronaut.security.fetch-metadata.rules.allow-simple-navigation` (default `true`) for symmetry.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/FetchMetadataRulesConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @return Whether to enable {@link SameOriginFetchMetadataRule} bean which allows requests whose Fetch Metadata identifies their initiator as the same origin.
+     */
+    boolean isAllowSameOrigin();
+
+    /**
+     * @return Whether to enable {@link SameSiteFetchMetadataRule} bean which allows requests whose Fetch Metadata identifies their initiator as the same site.
+     */
+    boolean isAllowSameSite();
```
dstepanov, line 61: Follow-on from the `SimpleTopLevelNavigationAndIframingFetchMetadataRule` comment: this interface exposes a toggle for five of the six built-in rules. Adding `isAllowSimpleNavigation()` here (plus the property and setter) would make the configuration surface complete and self-consistent.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/CrossOriginFetchMetadataRule.java
```diff
@@ -0,0 +1,101 @@
...
+
+    static boolean matchesMethod(CorsOriginConfiguration configuration, HttpRequest<?> request) {
+        HttpMethod requestMethod = request.getMethod();
+        HttpMethod methodToMatch = CorsUtil.isPreflightRequest(request)
+            ? request.getHeaders().getFirst(ACCESS_CONTROL_REQUEST_METHOD, HttpMethod.class).orElse(requestMethod)
+            : requestMethod;
+        List<HttpMethod> allowedMethods = configuration.getAllowedMethods();
+        return Objects.equals(allowedMethods, CorsOriginConfiguration.ANY_METHOD) ||
+            allowedMethods.contains(methodToMatch);
+    }
```
dstepanov, line 100: **This duplicates `CorsFilter`'s private validation logic and will drift.**

`matchesOrigin` and `matchesMethod` are near-verbatim reimplementations of micronaut-http-server's `CorsFilter` (5.1.12, lines 406-412, 423, 431-436) — right down to the `allowedOriginsRegex.isEmpty() && isAny(...)` subtlety, which is subtle enough that nobody will notice when it changes upstream.

Two consequences:

1. It already diverges — `CorsFilter` additionally applies `shouldDenyToPreventDriveByLocalhostAttack`, which this copy omits, so the Fetch Metadata exemption is slightly more permissive than the CORS filter it claims to stay aligned with.
2. Any future change to core's CORS semantics leaves this stale, silently and in the permissive direction.

Better options, roughly in order of preference: push a public helper into micronaut-core (`CorsUtil.isOriginAllowed(config, origin)` / `isMethodAllowed(...)`) and call it from here, or key the exemption off a request attribute that `CorsFilter` sets, rather than re-deriving the decision.

Minor, while you're in here: `Pattern.matches(regex, origin)` on line 81 recompiles the pattern on every cross-origin request.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/CrossOriginFetchMetadataRule.java
```diff
@@ -0,0 +1,101 @@
...
+ * permits the request origin and HTTP method.
+ *
+ * <p>Merely targeting a route annotated with
+ * {@link io.micronaut.http.server.cors.CrossOrigin CrossOrigin} is insufficient. The request
+ * must include an {@code Origin} header accepted by that route's CORS configuration. This keeps
+ * the Fetch Metadata exemption aligned with the validation performed by Micronaut's CORS
+ * filter. For a preflight request, the requested method is read from the
+ * {@code Access-Control-Request-Method} header.</p>
+ */
+@Requires(classes = { HttpRequest.class, CorsOriginConfiguration.class, CrossOriginUtil.class })
```
dstepanov, line 51: `CorsUtil` is used on line 94 but isn't listed in `@Requires(classes = ...)`. In practice all three classes ship in the same artifact so this can't actually blow up, but the list reads as if it's exhaustive — worth adding for consistency.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/FetchMetadataFilter.java
```diff
@@ -0,0 +1,102 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+/**
+ * Server filter that enforces the configured Fetch Metadata request-isolation policy.
+ */
+@Requires(property = FetchMetadataFilterConfigurationProperties.PROPERTY_ENABLED, value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
+@Requires(classes = ServerFilter.class)
+@ServerFilter("${" + FetchMetadataFilterConfigurationProperties.PREFIX + ".pattern:" + ServerFilter.MATCH_ALL_PATTERN + "}")
```
dstepanov, line 41: **An empty `pattern` silently disables the filter.**

This reads the raw property placeholder, and nothing in the module injects `FetchMetadataFilterConfiguration` — so neither `setPattern`'s empty-guard nor the `@NotBlank` on the field is in the request path.

Set `micronaut.security.fetch-metadata.filter.pattern: ""` (a plausible typo or a mis-substituted env var) and the placeholder resolves to the empty string: `@ServerFilter("")` matches nothing, the filter never runs, and there is no warning. Fail-open from a config typo is a rough failure mode for a security filter.

`FetchMetadataFilterConfigurationPropertiesTest.usesDocumentedDefaultsAndIgnoresEmptyPatterns` asserts the setter guard, which gives false confidence — that code path is never exercised in production.

Injecting the configuration bean and deriving the pattern from it would make the guard real.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/FetchMetadataFilterConfigurationProperties.java
```diff
@@ -0,0 +1,83 @@
...
+    /** Configuration prefix for the Fetch Metadata request filter. */
+    public static final String PREFIX = "micronaut.security.fetch-metadata.filter";
+    /** Property that enables or disables the Fetch Metadata request filter. */
+    public static final String PROPERTY_ENABLED = FetchMetadataFilterConfigurationProperties.PREFIX + ".enabled";
+    /** Default filter enablement. */
+    public static final boolean DEFAULT_ENABLED = true;
+    /** Default server-filter pattern. */
+    public static final String DEFAULT_PATTERN = "/**";
+
+    @NotBlank
```
dstepanov: `@NotBlank` is inert here: `@ConfigurationProperties` beans are only validated when the validation annotation processor is on the classpath, and the module only pulls in `api(mnValidation.validation)` (the API). Combined with the fact that the filter reads the property placeholder rather than this bean (see my comment on `FetchMetadataFilter`), this constraint never runs.

Either wire up validation properly or drop the annotation and the `jakarta.validation` dependency — right now it's an `api` dependency carried for one annotation that does nothing.

Unrelated nit on line 30: this class is package-private while `FetchMetadataRulesConfigurationProperties` is `public final`. Both are `@ConfigurationProperties` in the same module; worth picking one convention.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/FetchMetadataFilter.java
```diff
@@ -0,0 +1,102 @@
...
+                    LOG.trace("request {} {} rejected by rule {}", request.getMethod(), request.getPath(), rule.getClass().getSimpleName());
+                }
+                return forbidden();
+            }
+        }
+        if (LOG.isTraceEnabled()) {
+            LOG.trace("request {} {} rejected because no Fetch Metadata rule allowed it",
+                request.getMethod(), request.getPath());
+        }
+        return forbidden();
```
dstepanov: A 403 from a security filter is usually worth more than `TRACE`. As it stands an operator debugging a blocked client sees an unexplained 403 with nothing in the logs at default levels.

Suggest `DEBUG` for the rejection paths (keeping `TRACE` for the approvals), and including the actual `Sec-Fetch-Site` / `-Mode` / `-Dest` values in the message — that's the first thing anyone will want when triaging.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/FetchMetadataFilter.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private static MutableHttpResponse<Object> forbidden() {
```
dstepanov, line 108: The rejection is a bare `403` with no body and no extension point. Elsewhere in this repo rejections go through `RejectionHandler`, so applications can render a page, redirect, or add a header.

Even just a `@DefaultImplementation`-backed seam here would let applications customize the response without forking the filter.

security-fetch-metadata/src/main/java/io/micronaut/security/fetchmetadata/rules/FetchMetadataRulesConfigurationProperties.java
```diff
@@ -0,0 +1,124 @@
...
+        this.allowSameSite = allowSameSite;
+    }
+
+    @Override
+    public boolean isAllowNoFetchMetadata() {
+        return allowNoFetchMetadata;
+    }
+
+    /**
+     * Whether to enable {@link NoFetchMetadataRule} bean which allows requests without a complete set of Fetch Metadata headers.. Default value: {@value #DEFAULT_ALLOW_NO_FETCH_METADATA}
```
dstepanov: Typo: `headers..` — double period.

security-fetch-metadata/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+import io.micronaut.build.TestFramework
+
+plugins {
+    id("io.micronaut.build.internal.security-module")
+}
+dependencies {
+    api(mn.micronaut.http)
+    api(mnValidation.validation)
```
dstepanov: See my comment on `FetchMetadataFilterConfigurationProperties`: this `api` dependency exists solely for a `@NotBlank` that never gets evaluated. If the annotation goes, this can go with it.

src/main/docs/guide/fetchMetadata/fetchMetadataResourceIsolationPolicy.adoc
```diff
@@ -0,0 +1,50 @@
...
+|Same site
+|Allows `Sec-Fetch-Site: same-site`. Enable this only when every same-site origin is trusted.
+|Disabled
+
+|Browser initiated
+|Allows `Sec-Fetch-Site: none`, which represents requests explicitly initiated through browser UI, such as the address bar or a bookmark.
+|Enabled
+
+|Simple navigation
+|Allows `GET` requests with `Sec-Fetch-Mode: navigate`, except `object` and `embed` destinations.
```
dstepanov, line 33: The simple-navigation rule allows cross-site `GET` navigations with `frame`/`iframe` destinations — i.e. the default policy permits the application to be iframed cross-site. That matches the web.dev reference policy, so no complaint about the behaviour, but the table row ("except `object` and `embed`") makes it easy to read the opposite.

Worth one sentence noting that framing protection still requires CSP `frame-ancestors` or `X-Frame-Options`, so nobody assumes this module covers clickjacking.

dstepanov (commented):
Nice module — clean `Ordered` rule SPI, per-rule toggles, defaults that track the web.dev reference policy, and honest docs about the limitations. Functional coverage is good too.

Left inline comments. The two I'd most like to see addressed before merge are the fail-open on *unrecognized* `Sec-Fetch-*` values (`NoFetchMetadataRule`) and the missing toggle on `SimpleTopLevelNavigationAndIframingFetchMetadataRule`. The rest are smaller.

One thing I checked and it's fine: `@since 5.4.0` and `binaryCompatibility.enabledAfter("5.4.0")` both line up with `projectVersion=5.4.0-SNAPSHOT`.

PR #2273, opened 2026-08-25
Bypass authentication for anonymous static resources

### LINKED ISSUE #2267: Provide a way to avoid resolving the authentication for some URL pattern
### Feature description

We use basic JWT authentication with cookie enabled.
We noticed that, for static assets, the authentication is still loaded which due to our security layer do a query on database.

We do configure 

```yaml
micronaut:
  security:
    intercept-url-map:
      - pattern: "/ui/**"
        access: "isAnonymous()"
```

But even with that, when the cookie is added to  the asset request, the cookie authentication is performed.
This is because the Authentication object is resolved even for anonymous path so you can inject an optional Authentication object.

We would need a way to bypass authentication resolving entirely for some URL pattern.

sdelamo:
Add an opt-in security filter setting that skips authentication fetchers for unrouted GET and HEAD requests resolving to static resources with anonymous access.

Provide a replaceable bypass strategy, preserve authentication for routed and protected resources, and cover enabled, disabled, and custom implementations with tests.

Close: https://github.com/micronaut-projects/micronaut-security/issues/2267

### DIFF
--- a/security/src/main/java/io/micronaut/security/filters/DefaultStaticResourceAuthenticationBypass.java
+++ b/security/src/main/java/io/micronaut/security/filters/DefaultStaticResourceAuthenticationBypass.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.security.filters;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.http.HttpMethod;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.security.rules.ConfigurationInterceptUrlMapRule;
+import io.micronaut.web.router.RouteAttributes;
+import io.micronaut.web.router.RouteMatch;
+import io.micronaut.web.router.resource.StaticResourceResolver;
+import jakarta.inject.Singleton;
+
+/**
+ * Default static-resource-based implementation of {@link StaticResourceAuthenticationBypass}.
+ */
+@Internal
+@Requires(classes = { HttpRequest.class, StaticResourceResolver.class })
+@Requires(beans = { ConfigurationInterceptUrlMapRule.class, StaticResourceResolver.class })
+@Requires(missingBeans = StaticResourceAuthenticationBypass.class)
+@Singleton
+final class DefaultStaticResourceAuthenticationBypass implements StaticResourceAuthenticationBypass<HttpRequest<?>> {
+
+    private final StaticResourceResolver staticResourceResolver;
+    private final ConfigurationInterceptUrlMapRule interceptUrlMapRule;
+
+    DefaultStaticResourceAuthenticationBypass(StaticResourceResolver staticResourceResolver,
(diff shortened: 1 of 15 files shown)


dstepanov (commented):
Reviewed the implementation, tests and docs. The core design is sound, and I like that the bypass only skips the `AuthenticationFetcher`s — `checkRules` still runs the full `SecurityRule` chain with a `null` authentication, so an over-eager `shouldBypass` fails closed (401) rather than opening a hole. Opt-in flag, default `false`, and the docs spell out the exact preconditions plus the `WARNING` about downstream code seeing no `Authentication`. Test coverage is good: enabled (GET + HEAD), disabled, routed-takes-precedence, non-anonymous static resource, and a custom implementation.

A few non-blocking observations:

**1. `DefaultStaticResourceAuthenticationBypass#shouldBypass` — cheap check last**

```java
return staticResourceResolver.resolve(request.getUri().getPath()).isPresent()
    && interceptUrlMapRule.isAnonymous(request);
```

Both operands are side-effect free, but `StaticResourceResolver#resolve` does actual resource lookup while `isAnonymous` is an in-memory Ant-pattern match over a pre-built list. Swapping them means the resolve only happens for paths the intercept-url-map already marks anonymous, and requests that don't match any pattern (the common case in an app with a narrow `/assets/**` entry) never pay for it. The javadoc says to "resolve the resource before consulting the intercept URL map", but since a route match is already ruled out above, I don't see a correctness reason for that order — if there is one, it'd be worth stating explicitly.

Also worth noting: on a bypassed request the resource is resolved twice (once here, once by the static resource handling itself). Still a net win versus a DB round-trip, but it's not free.

**2. `ConfigurationInterceptUrlMapRule#isAnonymous` is `@Internal` but `public` on a public, user-extensible class**

It'll show up in the published API and in IDE completion for anyone extending `ConfigurationInterceptUrlMapRule`, and the default bypass depends on the concrete class rather than an abstraction. Since `findPattern` is already `protected` on `InterceptUrlMapRule`, an alternative is for the bypass to hold the pattern list / matching itself, or to expose this through a small internal interface. Minor, but new public surface in a security module is hard to walk back.

(Nit: `InterceptUrlMapRule#findPattern` is new `protected` API without an `@since` tag, unlike the other additions in this PR.)

**3. Naming of the extension point**

`StaticResourceAuthenticationBypass<T>` is generic over the request type and its contract (`boolean shouldBypass(T request)`) is entirely general — "static resource" is a property of the default implementation, not of the SPI. Once this ships as public API the name is fixed. Something like `AuthenticationBypass` / `AuthenticationResolutionBypass` would describe the contract, with the static-resource semantics living in the default bean and in the config key. Happy to defer if the intent is to keep the extension point deliberately narrow.

**4. Custom implementations still need the flag**

A user who registers their own `StaticResourceAuthenticationBypass` bean also has to set `micronaut.security.filter.static-resource-authentication-bypass=true` or it silently does nothing. It's documented, and the conservative default is defensible, but it is a footgun — providing the bean is already an explicit opt-in.

Related: any existing subclass calling the deprecated 3-arg `SecurityFilter` constructor gets `staticResourceAuthenticationBypassEnabled = false`, so the flag is silently inert for them. Probably fine given the constructor is now `forRemoval`, but it's a silent no-op rather than a warning.

**5. Scope vs. the linked issue**

#2267 asks for bypassing authentication resolution for a URL pattern. This delivers it only for unrouted requests that resolve to a static resource. That covers the reporter's `/ui/**` case if it is served via `micronaut.router.static-resources`, but not an anonymous *routed* endpoint. The `StaticResourceAuthenticationBypass` SPI is the escape hatch, so I think the narrow default is the right call — just flagging that the issue may not be fully closed for everyone who lands on it.

**6. Branch contains unrelated commits**

The diff carries ~10 files of CI/workflow/renovate/dependency changes (`.github/workflows/*`, `gradle/libs.versions.toml`, `MAINTAINING.md`, `settings.gradle`, `OpenIdClientFactorySpec`, `BuiltInEndpointsSpec`) from commits that also exist on `5.4.x` under different SHAs. It merges cleanly, but it makes the diff harder to read — a rebase onto current `5.4.x` would drop the noise.

Not approving yet only because `Java CI` is still in progress on `8971bab`; nothing above is blocking from my side once it's green.

---

graemerocher (approved):

PR #2293, opened 2026-09-14
Do not log authentication attribute values in SecurityFilter

sdelamo:
`SecurityFilter` logged every authentication attribute value at `DEBUG`. With the OAuth 2.0 module the attributes include the access, refresh and ID tokens, so enabling debug logging on the filter wrote bearer tokens to
the logs.

At `DEBUG` only the attribute keys are now logged. Values are still available at TRACE, but the values of the `accessToken`, `refreshToken` and `openIdToken` attributes, and of any attribute whose key ends in "token",
are redacted.

### Finding

**Problem.** `LOG.debug("Attributes: {}", ...) `prints every attribute value; with OAuth2 that includes access, refresh and ID tokens.

**Impact.** Enabling debug logging on the filter to troubleshoot writes bearer tokens to logs.


security/src/main/java/io/micronaut/security/authentication/AuthenticationUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods for {@link Authentication}.
+ *
+ * @since 5.4.0
+ */
+@Internal
+public final class AuthenticationUtils {
```
graemerocher: seems like the methods of this class should go into https://github.com/micronaut-projects/micronaut-security/blob/5.4.x/security/src/main/java/io/micronaut/security/utils/LoggingUtils.java since this adds a whole new generically named type with limited scope which may be confusing

graemerocher (changes requested):

graemerocher (approved):

security/src/main/java/io/micronaut/security/authentication/AuthenticationUtils.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-security/pull/2293/commits/add971275b2965ab7e016a306ffeb83200b371f2

PR #2295, opened 2026-09-14
Handle null attribute values in introspection responses

sdelamo:
`Authentication.build` accepts arbitrary attribute maps and, for example, the
OAuth2 authentication mapper can store a `refreshToken` attribute with a null
value. `DefaultIntrospectionProcessor` resolved attributes with
`containsKey(k) ? attributes.get(k).toString() : `..., so an attribute that was
present but null threw a NullPointerException and /token_info answered 500.

Resolve every string attribute through a null-safe attributeAsString helper
that treats a null value the same as an absent key, return an empty Optional
from secondsSinceEpochOfAttribute for null exp/iat/nbf values, and omit
null-valued extensions from the response instead of serialising them.

### Finding

**Problem.** Every resolver uses `containsKey(k) ? attributes.get(k).toString() : `..., and secondsSinceEpochOfAttribute falls through to obj.toString(). Authentication.build accepts arbitrary maps.

**Impact.** An attribute present with a null value (see L41 for a real producer) throws NullPointerException from /token_info.


graemerocher:
The diagnosis is right and the `attributeAsString` helper is the correct shape for the fix — it collapses eight copies of the same `containsKey`/`toString` pattern and makes "present but null" behave identically to "absent" everywhere. A few things worth addressing before merge.

**1. There's a second bug fixed here that isn't mentioned.** In `resolveExtensions`, an attribute whose key equals `tokenConfiguration.getRolesName()` and whose value is null previously made `extensions.containsKey(rolesName)` return `true`, so the `authentication.getRoles()` fallback never ran and the response advertised `"roles": null` instead of the actual roles. Skipping null values fixes that too, which is a real (if narrow) change to the response body. Worth a line in the PR description so it lands in the release notes rather than surprising someone.

**2. `attributeAsString` should probably be `protected`.** Every sibling resolver (`resolveScope`, `resolveClientId`, `resolveSub`, …) is `protected` precisely so subclasses can override them, and `DefaultIntrospectionProcessor` is a public class built around that extension pattern. A subclass overriding `resolveIssuer` to read a differently-named claim now has to re-implement the null-safe lookup by hand. Making the helper `protected` costs nothing and keeps the extension story coherent.

**3. `!json.contains('null')` is too broad an assertion.** It passes today, but it fails for reasons unrelated to the bug the moment any attribute value contains that substring — a username like `nullable`, a scope named `annullable`, an email at a domain containing it. Asserting on the specific keys that must be absent (`!json.contains('"refreshToken"')`, `!json.contains('"scope"')`) keeps the test pinned to what it's actually verifying.

**4. Minor:** `resolveExtensions` iterates `keySet()` and then calls `getAttributes().get(k)` for each key. Now that the body needs the value anyway, `entrySet()` drops a map lookup per attribute. Pre-existing, so ignore if you'd rather keep the diff tight.

The `secondsSinceEpochOfAttribute` change reads correctly — hoisting the `get` and null-checking it is exactly equivalent to the old `containsKey` guard except for the null case, and the javadoc update matches.

---

graemerocher (approved):

PR #2301, opened 2026-09-14
Use UTF-8 consistently for HMAC and Base64 byte conversions

sdelamo:
Addresses audit finding L10 (Low): HMacUtils and the CSRF generator mix UTF-8 with the platform default charset.

HMacUtils encoded the key with UTF-8 but the data with the platform
default charset, and DefaultCsrfTokenGenerator Base64-encoded the session
id with the platform default charset. HMACs and CSRF tokens therefore
differed across JVMs with different file.encoding whenever the input
contained non-ASCII characters.

- HMacUtils: hash the data bytes as UTF-8.
- DefaultCsrfTokenGenerator: Base64-encode the session id as UTF-8.
- KeysController: decode the JSON bytes produced by JsonMapper as UTF-8
  instead of the platform default.
- Add tests asserting that a non-ASCII message / session id produces the
  value computed with an explicit UTF-8 Mac.

### Finding

**Problem.** The HMAC key uses StandardCharsets.UTF_8 while the data uses getBytes() with the platform default; the CSRF generator Base64-encodes the session id with getBytes() too.

**Impact.** HMACs differ across JVMs with different file.encoding when the data contains non-ASCII characters.

### Notes

- Also fixed KeysController charset; scratchpad commit-msg collision, verify subjects at end.
- The commit is not signed.


graemerocher:
The change is correct and, as far as I can tell, complete — I grepped the main sources and these three are the only unqualified `getBytes()` / `new String(byte[])` sites across all modules, so there's no fourth one lurking.

Two substantive points.

**1. This is not a breaking change for the shipped CSRF flow, and the PR should say so.** The "HMACs differ across JVMs" framing reads as though existing deployments could start rejecting tokens after upgrade. They can't, for the built-in path: `hmacMessagePayload` builds its message entirely out of Base64 output, decimal lengths and `!` separators — all ASCII — and the session id is itself Base64-encoded before it ever reaches `base64EncodedHmac`. So the bytes fed to the `Mac` are identical under every charset today, and the only in-tree caller of `HMacUtils` is `DefaultCsrfTokenGenerator`.

The change does matter for two cases worth naming in the release note:
- a custom `SessionIdResolver` returning a non-ASCII session id, and
- third-party callers of `HMacUtils`, which is public API.

For those, previously-issued values stop validating on a JVM where `file.encoding` isn't UTF-8. That's the right trade, but it's a much smaller blast radius than the description implies, and saying so explicitly will save a reviewer the same ten minutes it cost me.

**2. Neither new test can fail on CI.** This is the one I'd actually want addressed. Both tests compute `expected` with an explicit UTF-8 `Mac` and compare it against a hardcoded literal that was itself produced under UTF-8. On any JVM whose default charset is already UTF-8 — which is every JVM the build runs on, given Java 18+ fixes `file.encoding` to UTF-8 per JEP 400 — reverting `getBytes(StandardCharsets.UTF_8)` back to `getBytes()` leaves both tests green. They document the intended behaviour but don't guard the regression.

If you want them to bite, the test JVM needs a non-UTF-8 default charset — e.g. a small `Test` task with `systemProperty("file.encoding", "ISO-8859-1")` (or `-Dsun.jnu.encoding`) running just these two classes. If that's more machinery than it's worth, that's a reasonable call, but then I'd drop the "it must not change with the JVM's file.encoding" comments, since they claim a guarantee the tests aren't providing.

Minor, take it or leave it: in both tests the `assertEquals(literal, expected)` line is the useful canary and `assertEquals(expected, actual)` is the real assertion — comparing the production value against the literal directly would express the same thing in one step.

---

PR #2346, opened 2026-09-14
Do not log CSRF token values

sdelamo:
`RepositoryCsrfTokenValidator` logged the request token at WARN when it did not contain the HMAC/random separator. That branch is only reached after the request token already matched the stored repository token, so
the live token was written to the log, and the attacker-supplied value was a log-flooding and log-injection vector.

Log a fixed message at DEBUG instead, without the token value. Add a test that attaches a logback ListAppender to the validator's logger and asserts the token is absent from all messages and nothing is logged at WARN or above.

### Finding

**Problem.** The branch logging `csrfTokenInRequest` is only reached after equals matched the repository token, so the value logged is the live stored token.

**Impact.** Secret material in production logs.


graemerocher:
I verified the finding and it holds: `validateCsrfToken` only reaches `validateHmac` after `csrfTokenInRepository.equals(csrfTokenInRequest)` has already matched, so the value going to `LOG.warn` was necessarily the live stored token. Removing it is right.

**The rationale needs one correction, and it changes the conclusion.** The PR says the attacker-supplied value was "a log-flooding and log-injection vector". By the same reasoning that makes this a secret leak, it isn't: an attacker can't reach this branch without already holding a token equal to the stored one. So flooding and injection aren't available here, and the only real problem was the secret in the log.

That matters because it undercuts the move to DEBUG. Now that the value is gone, there's nothing unsafe about WARN, and the condition being reported is genuinely alarming — it means a `CsrfTokenRepository` is handing back a token that `CsrfHmacTokenGenerator` didn't produce, i.e. a misconfiguration. That's exactly what an operator wants surfaced. I'd keep WARN and just drop the value; DEBUG buries a real configuration fault.

**The message points at the wrong party.** "CSRF token rejected: expected two dot-separated segments" reads as though a malformed request was rejected. But per the above, the request token matched the repository token, so what's malformed is the token *in the repository*. Something like `CSRF token in repository is not in the expected hmac.random format` is the actionable diagnosis — it tells the operator where to look instead of sending them off to inspect client requests that are fine.

**Minor:** "dot-separated" hardcodes `CsrfHmacTokenGenerator.HMAC_RANDOM_SEPARATOR` in prose while the code beside it reads the constant. If the separator ever changes, the message silently goes stale.

---

**Separate issue in the same method, worth raising while you're here.** `validateCsrfToken` does:

```java
if (csrfTokenInRepository.equals(csrfTokenInRequest) && validateHmac(request, csrfTokenInRequest)) {
```

`String.equals` short-circuits on the first differing character, so this compares a stored secret against attacker-controlled input in variable time — and it's the *first* comparison, gating everything else. `validateHmac` then goes to the trouble of using `MessageDigest.isEqual` for the HMAC, which is good practice made largely moot by the comparison upstream of it. `MessageDigest.isEqual` on the UTF-8 bytes would make the two consistent. Happy for that to be a follow-up rather than scope creep here, but it's the more interesting problem in this file.

**CI:** `SonarCloud Code Analysis` is red on the head commit; the rest is green.

The test is solid — asserting both that the value is absent *and* that nothing lands at WARN or above is the right pair of assertions, and it'll catch a regression that only half-fixes this.

---

graemerocher (approved):

PR #2352, opened 2026-09-14
Exclude password attributes from LDAP search results by default

sdelamo:
Addresses audit finding L62 (Low): User search returns every attribute by default, including userPassword hashes when readable.

When `micronaut.security.ldap.*.search.attributes` is not set, the user
search requests every attribute the manager account can read, including
`userPassword` hashes, which were then handed to the
ContextAuthenticationMapper and could reach trace logging.

Keep `attributes: null` meaning "all attributes", but strip a configurable
list of sensitive attributes from the returned Attributes before they are
wrapped in LdapSearchResult:

- Add `micronaut.security.ldap.*.search.excluded-attributes`
  (List<String>, default [userPassword, unicodePwd], case-insensitive).
  An empty list disables the exclusion.
- Add `SearchSettings.getExcludedAttributes()` (default empty, so custom
  SearchSettings implementations keep their previous behaviour) and
  implement it in SearchPropertiesSearchSettings.
- DefaultLdapSearchService strips the excluded attributes in a new
  `createResults(results, excludedAttributes)` overload; the existing
  `createResults(results)` is kept and delegates with no exclusions.
- Document the behaviour and recommend an explicit minimal `attributes`
  list in the LDAP configuration guide.
- Add LdapSearchExcludedAttributesSpec covering the default, a cleared
  list, an extended list, an explicit attributes list, and an excluded
  attribute requested explicitly.

### Finding

**Problem.** SearchConfiguration.attributes defaults to null and setReturningAttributes(null) requests all attributes.

**Impact.** If the manager account can read userPassword, hashes are pulled into memory and handed to ContextAuthenticationMapper and trace logging. Depends on directory ACLs.

### Notes

- Property search.excluded-attributes.
- The commit is not signed.


graemerocher:
Worthwhile change and the docs addition is good — the "set `attributes` explicitly to the minimal set" recommendation is the real fix and I'm glad it's there rather than leaving the deny-list as the whole answer. Two things I think need addressing before merge, plus some smaller ones.

## 1. Existing subclasses of `DefaultLdapSearchService` silently stop working

`createResults(NamingEnumeration)` is `protected` and has been the extension point since 1.0. After this change `search()` does:

```java
if (CollectionUtils.isEmpty(excludedAttributes)) {
    return createResults(results);
}
return createResults(results, excludedAttributes);
```

With the new default (`[userPassword, unicodePwd]`) that second branch is the one that runs, so anyone who has subclassed `DefaultLdapSearchService` and overridden `createResults(NamingEnumeration)` — to attach extra data to `LdapSearchResult`, filter entries, whatever — finds their override is no longer called. No compile error, no config change on their part, just their code quietly dropping out of the path on upgrade.

The conditional dispatch makes it worse: whether the override runs now depends on whether the operator cleared `excluded-attributes`, so the same code behaves differently across two deployments of the same app.

I'd have `search()` call the two-arg overload unconditionally and make the one-arg version delegate to it (`createResults(results)` → `createResults(results, Collections.emptyList())`), which it already does. That keeps source compatibility, makes behaviour independent of configuration, and if you deprecate the one-arg overload at the same time, subclassers get a visible signal instead of a silent one.

## 2. An explicit `attributes` list is overridden without a word

`"an excluded attribute is stripped even when requested explicitly"` codifies that if an operator writes:

```yaml
search:
  attributes: [cn, userPassword]
```

the `userPassword` they deliberately asked for is dropped, with no log line and nothing in the config to hint at why. An explicit allow-list is a stronger and more specific signal than an inherited default deny-list, so having the default win silently is the surprising direction. I'd either let an explicit `attributes` entry take precedence, or keep the strip and log a WARN once naming the conflicting attribute — silent is the one option I'd rule out, because the failure mode is someone spending an afternoon on "why is my attribute missing".

The same edge bites `groups.filter-attribute`: excluding an attribute that the group filter reads breaks group resolution with no diagnostic at all. Your new doc NOTE warns about this for `search.attributes` but not for `excluded-attributes`, and the latter is now the one that's non-empty by default.

## 3. "out of memory" overstates what this does

The PR body and the guide both frame this as keeping hashes out of memory. It doesn't — the server still returns them, they arrive in the JNDI response and sit in the `Attributes` object, and this code removes them a moment later. What it genuinely prevents is hashes reaching `ContextAuthenticationMapper` and trace logging, which is a real defense-in-depth win and reason enough for the change. But the doc paragraph should say "out of your application code" rather than implying they never land in the process, so nobody builds a compliance argument on top of a guarantee that isn't being made.

## 4. Custom `SearchSettings` implementations get no protection

`getExcludedAttributes()` defaults to empty, so the new default only covers the properties-driven path. The PR body calls this out as deliberate backwards-compatibility and that's defensible — a non-empty interface default would change behaviour for implementations that never opted in. But it should be one line in the guide, so someone running a custom `SearchSettings` doesn't read the new docs and assume they're covered.

## 5. Smaller items

- **`removeExcludedAttributes` mutates in place.** It calls `attributes.remove(id)` on the instance from `result.getAttributes()` and returns that same instance, but the javadoc says "The attributes without the excluded ones", which reads like a copy. It's `protected`, so subclasses will assume the wrong contract. Worth stating the mutation explicitly.
- **Release note needed.** An app that verifies against `userPassword` itself breaks silently on upgrade. Uncommon, but the `excluded-attributes: []` escape hatch should be documented where someone hitting it will find it.
- **`getExcludedAttributes()` returns the internal list and `setExcludedAttributes` stores the caller's.** Not a practical risk for config binding, but `List.copyOf` in the setter would be tidier for a public config class.

The spec coverage is good — default, cleared, extended, explicit list, and the explicit-but-excluded case are exactly the five you'd want, and I'd only change what the last one asserts rather than that it exists.

---

PR #2353, opened 2026-09-14
Copy OracleJsonObject attributes before merging end-user context

sdelamo:
Addresses audit finding L63 (Low): Attribute merge mutates OracleJsonObject instances taken from Authentication.getAttributes.

DefaultAttributesFetcher stored the live OracleJsonObject instances taken
from Authentication.getAttributes() and merge() called putAll() on them
when a later attribute contained the same END USER CONTEXT name. With
attributeNames=a,b the object stored in the Authentication for "a" was
mutated with the keys of "b": subsequent getConnection calls merged
already-merged data and any other consumer of the Authentication saw the
changed attributes.

The fetcher now copies every OracleJsonObject it takes from an
Authentication via OracleJsonFactory.createObject(OracleJsonObject), and
merge() creates a new JSON object instead of mutating the existing one.

### Finding

**Problem.** parseAttributes stores the live object when an attribute is already an OracleJsonObject, and merge calls existingJson.putAll(currentJson) on it when a later map has the same context name.

**Impact.** With attributeNames=a,b both containing scott.user_info, the object stored in the Authentication for a is mutated with b's keys; subsequent getConnection calls and other consumers see changed attributes.

### Notes

- The commit is not signed.


graemerocher:
Good catch, and the fix is complete — I traced both routes by which a live object can escape an `Authentication` into the fetcher's maps:

- `toAttributes(String, OracleJsonObject)` → `parseAttributes(String, OracleJsonObject)`, now copying per context value
- `toAttributes(String, Map)` → `toAttributeValues(...)`, now copying the `OracleJsonObject` and `OracleJsonValue` branches

The `CharSequence` and `Map` branches already produced fresh objects via `parseJsonObject`/`toJsonObject`, so there's no remaining path that stores a caller-owned instance. The test is well chosen too: running `fetchAttributes` twice catches both the mutation itself and the "already-merged data gets re-merged" symptom that would otherwise look like a passing test.

Three things:

**1. Is `createObject(OracleJsonObject)` a deep or a shallow copy?** It matters for the javadoc you've added on `copyOf` ("Creates a mutable copy"). `merge` only ever does a top-level `putAll`, so a shallow copy is sufficient for the bug being fixed — but if it is shallow, a nested object inside a context value is still shared with the `Authentication`, and the next person who adds a nested merge reintroduces exactly this bug against a javadoc that reads as though it can't happen. Stating the depth explicitly in the comment would close that off either way.

**2. The `copyOf` inside the merge function is now redundant.** By the time `merged.merge(...)` runs, every value in both maps came from `parseAttributes` or `toAttributeValues` and is already fetcher-owned, so `existingJson` can be mutated safely. It's harmless as belt-and-braces, but it allocates a fresh object per colliding key on every connection checkout. Either drop it and mutate `existingJson` directly, or keep it and say in a comment that it's defensive rather than load-bearing — right now it reads as load-bearing, which will stop anyone from ever simplifying it.

**3. Small double-copy on the parse path.** `parseAttributes(String, String)` calls `parseJsonObject`, which builds a fresh object, and then hands it to `parseAttributes(String, OracleJsonObject)`, which copies every value again. That's an extra copy per context name for the `endUserContextAttributes` parameter and for `CharSequence` authentication attributes, on a per-checkout path. Pushing the `copyOf` into only the callers that receive foreign objects would avoid it, at the cost of a slightly less obviously-safe invariant — your call on whether that trade is worth making.

Also worth noting for the changelog: `merge` went from `static` to an instance method. No API impact since it's private, just flagging it so it doesn't look accidental in review.

---

graemerocher (approved):

PR #2354, opened 2026-09-14
Log OJDBC end-user context attribute values only at TRACE

sdelamo:
Addresses audit finding L64 (Low): Debug log dumps all resolved end-user-context attributes per connection checkout.

The DEBUG log written by MicronautEndUserSecurityContextProvider on every
connection checkout dumped the full END USER CONTEXT attribute map, which is
derived from JWT claims and authentication attributes and may contain
personal data. At DEBUG only the data roles, the attribute names and their
counts are logged now; the full attribute values remain available at TRACE.

Adds tests that attach a logback ListAppender at DEBUG and TRACE and assert
the attribute value is absent at DEBUG (while the attribute name is present)
and present at TRACE.

### Finding

**Problem.** At DEBUG the full attribute map derived from JWT claims and authentication attributes is written per checkout.

**Impact.** PII in staging logs where DEBUG is commonly enabled.

### Notes

- The commit is not signed.


graemerocher:
Agreed with the premise — a per-checkout DEBUG line carrying the full attribute map is the kind of thing that quietly fills a staging log with claim-derived personal data, and moving the values to TRACE is the right call. A few notes.

**1. The message logs both the count and the collection.** It now renders as:

```
GET /foo - end user security context resolved with 1 data roles [REPORTING] and 1 attributes [hr.employee]
```

The counts are redundant when the collections themselves are printed, and the interleaving reads awkwardly. Pick one: either counts only (`with 1 data roles and 1 attributes`), which is the more defensible choice if the worry is what a log line reveals about an individual, or collections only, which is what actually helps someone debugging a missing context attribute.

**2. Data role names are still logged in full at DEBUG.** That's probably fine — roles are authorization data rather than personal data — but it is a per-user value, and the PR's own rationale ("PII in staging logs where DEBUG is commonly enabled") applies to a role set that identifies someone's department or clearance just as much as it applies to an attribute value. Worth being a deliberate decision rather than the one that fell out of the diff.

**3. `traceLoggingIncludesAttributeValues` is asserting on a pre-existing log statement.** Nothing in this PR emits attribute values at TRACE — the assertion passes because of the `resolved the attributes {}` trace line that was already in `getEndUserSecurityContext`. That's fine and the test is still meaningful (it pins the TRACE-side contract the PR description promises), but a one-line comment in the test saying which statement it's covering would save the next reader the hunt, and would fail loudly if that trace line were ever removed for the wrong reason.

**4. CI:** `SonarCloud Code Analysis` is red on the head commit. Everything else is green, so it's probably a gate rather than a test, but worth a look before merge.

The test approach itself — `ListAppender` at both levels, asserting the name is present and the value absent at DEBUG — is the right way to pin this, and restoring the original level in a `finally` avoids leaking state into sibling tests.

---

graemerocher (approved):

